PR #13, opened 2016-05-19
precompile if available

rdeits:
I've been using this for several days with no issues, and I can't see any reason why precompilation wouldn't work for this module.


awbsmith:
Approving this would help out Rotations.jl

SimonDanisch:
Seems reasonable, thanks! :)

PR #14, opened 2016-05-25
angle is now numerically stable

jw3126:
The current `angle` method (called `angle_fast` in the following) is numerically instable:

```
using Quaternions

angle_stable(q::Quaternion) = 2*atan2(√(q.v1^2 + q.v2^2 + q.v3^2), q.s)
angle_fast(q::Quaternion) = 2 * acos(real(normalize(q)))

ax = randn(3)
θ = 1e-8
q = qrotation(ax, θ)
@show θ
@show angle_stable(q)
@show angle_fast(q)
```
```
θ = 1.0e-8
angle_stable(q) = 1.0e-8
angle_fast(q) = 0.0
```

I replaced it with `angle_stable`, which is s̶̶l̶̶o̶̶w̶̶e̶̶r̶ as fast:

```
using BenchmarkTools

q = Quaternion(randn(4)...)
println(median(@benchmark angle_stable($q)))
println(median(@benchmark angle_fast($q)))

BenchmarkTools.TrialEstimate: 
  time:             21.00 ns
  gctime:           0.00 ns (0.00%)
  memory:           0.00 bytes
  allocs:           0
  time tolerance:   5.00%
  memory tolerance: 1.00%
BenchmarkTools.TrialEstimate: 
  time:             23.00 ns
  gctime:           0.00 ns (0.00%)
  memory:           0.00 bytes
  allocs:           0
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

### DIFF
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -93,7 +93,7 @@ end
 
 angleaxis(q::Quaternion) = angle(q), axis(q)
 
-angle(q::Quaternion) = 2 * acos(real(normalize(q)))
+angle(q::Quaternion) = 2*atan2(√(q.v1^2 + q.v2^2 + q.v3^2), q.s)
 
 function axis(q::Quaternion)
     q = normalize(q)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,34 +1,53 @@
 using Base.Test
 using Quaternions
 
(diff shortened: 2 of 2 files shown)


SimonDanisch:
This seems to be redundant to #15 ?

jw3126:
Right!

PR #15, opened 2016-05-27
Tons of tests

jw3126:
Now almost all functions in Quaternion.jl are tested! There are also very few minor fixes and performance boosts. I did not care about julia 0.3, do we still need to support 0.3? If so I make things compatible, otherwise I would remove that requirement.


SimonDanisch:
Awesome! Thanks a lot!
We do not care about 0.3 anymore, see #17

PR #17, opened 2016-10-04
Turn on precompilation and remove support for Julia 0.3

timholy:

PR #19, opened 2016-12-01
Drop Julia 0.3 support

rdeits:
This is a rebased version of #17. 

Note: this also turns OFF osx builds on Travis. I'm happy to revert that change if you like, but I don't think there's much benefit to testing pure-julia packages like this on OSX as well as linux (and Travis's OSX machines are a relatively scarce resource). The benefit (in addition to being kind to Travis's servers) should be faster build times.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ os:
   - linux
   - osx
 julia:
-  - 0.3
   - 0.4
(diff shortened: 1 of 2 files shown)


rdeits:
@timholy would you be willing to review and merge this?

PR #22, opened 2017-05-31
Add a method to compute quaternion from direction cosine matrix

adambrewster:
rotationmatrix needs an inverse.  I don't know of qrotation is the right function.  I've usually called this function `dcm2q`, but `qrotation` seems close.  This PR includes the new method, a link to a paper describing the math, and tests that confirm `qrotation(rotationmatrix(q), q) ≈ q`.


SimonDanisch:
Cool, thank you! Guess everyone has overseen this PR ;)

PR #26, opened 2018-06-13
added ispure method for quaternion.

adamnemecek:


hyrodium:
Thanks for this PR! But I'll close this PR for the following reasons:

* This PR provides not just one feature.
    * This PR includes `rand`, `isfinite` etc. which are already solved in other PRs.
* I don't understand the practical benefit for `pure` and `ispure` functions.
    * We don't have `pure` function in `Base` which returns from `a+bi` to `bi`.

If you still need the functions, please open another PR with some descriptions.

PR #31, opened 2019-01-17
add float function

dkarrasch:
This is needed for `norm(::Quaternion)` and, for instance, for approximate matrix comparisons on Julia v1.2. I haven't added any tests, since there are currently no conversion tests.

Edit: BTW, are all those `Compat` things still necessary? I haven't seen that for a while.
Edit2: Oh, is that because Quaternions.jl is supposed to work for Julia v0.6 to nightly?


dkarrasch:
Bump.

KristofferC:
Fixed in #32

PR #32, opened 2019-05-17
Updates for Julia 1.0, 1.1 and 1.2

fredrikekre:


dkarrasch:
Not sure if, for completeness, also `Octonion` and the `Dual`s need a `float` function?

KristofferC:
I think that would be a good idea.

PR #35, opened 2019-08-11
Accept inputting `AbstractVector` and `AbstractMatrix`.

tianrluo:
Replaced `Vector`/`Matrix` with `AbstractVector`/`AbstractMatrix` to support creating quaternions from `view`s.

```julia-repl
julia> using Quaternions
julia> U, Φ = randn(5,3,4), randn(5,1,4);
julia> u, ϕ = view(U, 1,:,1), view(Φ, 1,:,1)[1];
julia> q = qrotation(u, ϕ)
```

------------
The `DualNumbers.jl` doesn't treat a 0-dimensional `AbstractArray` as a `Number`, hence the demo above has `ϕ = view(Φ, 1,:,1)[1]`.

### DIFF
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -17,8 +17,8 @@ Quaternion(s::Real, v1::Real, v2::Real, v3::Real, n::Bool = false) =
     Quaternion(promote(s, v1, v2, v3)..., n)
 Quaternion(x::Real) = Quaternion(x, zero(x), zero(x), zero(x), abs(x) == one(x))
 Quaternion(z::Complex) = Quaternion(z.re, z.im, zero(z.re), zero(z.re), abs(z) == one(z.re))
-Quaternion(s::Real, a::Vector) = Quaternion(s, a[1], a[2], a[3])
-Quaternion(a::Vector) = Quaternion(0, a[1], a[2], a[3])
+Quaternion(s::Real, a::AbstractVector) = Quaternion(s, a[1], a[2], a[3])
+Quaternion(a::AbstractVector) = Quaternion(0, a[1], a[2], a[3])
 
 convert(::Type{Quaternion{T}}, x::Real) where {T} = Quaternion(convert(T, x))
 convert(::Type{Quaternion{T}}, z::Complex) where {T} = Quaternion(convert(Complex{T}, z))
@@ -287,7 +287,7 @@ end
 
 ## Rotations
 
-function qrotation(axis::Vector{T}, theta) where {T <: Real}
+function qrotation(axis::AbstractVector{T}, theta) where {T <: Real}
     if length(axis) != 3
         error("Must be a 3-vector")
     end
@@ -302,7 +302,7 @@ function qrotation(axis::Vector{T}, theta) where {T <: Real}
 end
 
 # Variant of the above where norm(rotvec) encodes theta
-function qrotation(rotvec::Vector{T}) where {T <: Real}
+function qrotation(rotvec::AbstractVector{T}) where {T <: Real}
     if length(rotvec) != 3
         error("Must be a 3-vector")
     end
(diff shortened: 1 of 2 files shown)


sethaxen (commented):
Thanks for the PR! Can you add a test using `view`?

tianrluo:
Hey @sethaxen, I have added a simple test to the PR.
It is only partial regarding the change, but I haven't used `julia` in a long while, and am not sure how to modify the test in a simple way to cover the changed constructor: `Quaternion(a::AbstractVector)`.
It would be great if you could take a look into these!

test/test_Quaternion.jl
```diff
@@ -29,6 +29,9 @@ let # test rotations
     @test qrotation([0, 0, 0], 1.0) == Quaternion(1.0) # a zero axis should act like zero rotation
     @test qrotation([1, 0, 0], 0.0) == Quaternion(1.0)
     @test qrotation([0, 0, 0]) == Quaternion(1.0)
+    qx = qrotation(view([1, 0, 0], :), view([pi / 4], :)[1])
```
hyrodium, line 32: I think we don't need `view` in the second argument.
```suggestion
    qx = qrotation(view([1, 0, 0], :), pi / 4)
```

src/Quaternion.jl
```diff
@@ -13,8 +13,8 @@ Quaternion(s::Real, v1::Real, v2::Real, v3::Real, n::Bool = false) =
     Quaternion(promote(s, v1, v2, v3)..., n)
 Quaternion(x::Real) = Quaternion(x, zero(x), zero(x), zero(x), abs(x) == one(x))
 Quaternion(z::Complex) = Quaternion(z.re, z.im, zero(z.re), zero(z.re), abs(z) == one(z.re))
-Quaternion(s::Real, a::Vector) = Quaternion(s, a[1], a[2], a[3])
-Quaternion(a::Vector) = Quaternion(0, a[1], a[2], a[3])
+Quaternion(s::Real, a::AbstractVector) = Quaternion(s, a[1], a[2], a[3])
+Quaternion(a::AbstractVector) = Quaternion(0, a[1], a[2], a[3])
```
hyrodium, line 17: If we add tests here, it would be like this:
```julia
@test Quaternion(1, [1,2,3]) === Quaternion(1,1,2,3)
@test Quaternion([1,2,3]) === Quaternion(0,1,2,3)
```
Personally, I feel this constructor is not useful because:

* It doesn't check the length of the argument.
* Accepting `Vector` makes performance degression. (cf. #61 )

So. I'd like to remove them in the next breaking release, but the removing should be done in another PR.

sethaxen: > So. I'd like to remove them in the next breaking release, but the removing should be done in another PR.

Agreed. I think we should go further and make it so that every function that currently returns a vector returns a `Tuple` instead, and every function that accepts a vector can also accept a tuple. But yes, outside of this PR.

tianrluo: Would you recommend adding
```julia
@test Quaternion(1, [1,2,3]) === Quaternion(1,1,2,3)
@test Quaternion([1,2,3]) === Quaternion(0,1,2,3)
```
to the test, or just leave it since there will be a future PR with a different constructor?
(Sry I wasn't sure if you are waiting for my response, or if this conversation is considered resolved)

hyrodium: Sorry for the ambiguous comment. Yes, it seems better to have the tests I suggested. The tests will be useful when we will change the constructor behavior.

tianrluo: Thanks for clarifying! (and sorry about the late reply)
I have rebased the PR branch and included the test, but since the test files have been largely changed, I am not sure it would match what the team was looking for.
Let me know if you have further comments, thanks!

hyrodium (approved):
LGTM

Issue #29, opened 2018-08-29
Reexport Base.imag?


jiahao:
Currently, `Base.imag` is [not imported and re-exported](https://github.com/JuliaGeometry/Quaternions.jl/blob/master/src/Quaternions.jl#L8), which causes shadowing errors when intermixing `Complex` and `Quaternion` numbers in arithmetic computations. Since `im^2 == -1` in both algebras, the algebras themselves are not in conflict, and the analogous mixed-algebra computations should be similar to `Real * Complex` in `Base`. At least on paper, I don't foresee any problems in allowing `Quaternions.imag` to extend `Base.imag`.

JeffBezanson:
The `imag` function here returns a vector of the 3 imaginary components. Does that really implement the `imag` function? Doesn't seem right to me.

chakravala:
If you'd like to gain a better understanding geometric or algebraic understanding of imaginary units of quaternions and complex values, I recommend you play around with it on [Grasmann.jl](https://gitbub.com/chakravala/Grassmann.jl) a bit.

```Julia

julia> using Grassmann

julia> i,j,k = hyperplanes(ℝ^3)
3-element Array{SValue{⟨+++⟩,2,B,Int64} where B,1}:
 -1v₂₃
 1v₁₃ 
 -1v₁₂

julia> (ℝ^3)(I) # this is the 3D pseudoscalar
1v₁₂₃

julia> (ℝ^3)(I)^2 # squares to -1 also
-1v
```
In the quaternion algebra, there is an `i,j,k` which are all **planes** which square to `-1` but there is another capital `I` which is usually called the **volume** element which is a single imaginary unit.

Quaternions are generated by a 3D vector space, but complex numbers are generated by a 2D vector space combining a scalar value with a pseudoscalar value, i.e the volume element `V(I)`.
```Julia
julia> (ℝ^2)(I) # this is the 2D pseudoscalar
1v₁₂

julia> (ℝ^2)(I)^2 # squares to -1 as imaginary unit
-1v
```
As you can see, the pseudoscalar `V(I)` in 2D is `v12` while in 3D it is `v123`.

It so happens that the pseudoscalar `v12` from 2D is equal to `i` from the `i,j,k` of quaternions also.

Thus, you could either:
- equate `im` with `v12`, the bivector plane element, which would require a sign change for 3D!
- equate `im` with `V(I)`, the pseudoscalar element, which is more natural for `n`-dimensions

[...]

PR #40, opened 2020-07-11
Fix conversion of `Complex` to `Quaternion`/`Octonion`

dkarrasch:
Not sure if this addresses #29, but at least it allows the `Base` promotion to work. This method was previously not tested, and probably not working.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Quaternions"
 uuid = "94ee1d12-ae83-5a48-8b1c-48b8ff168ae0"
-version = "0.4.1"
+version = "0.4.2"
 
 [deps]
 DualNumbers = "fa6b7ba4-c1ee-5f82-b5fc-ecf0adba8f74"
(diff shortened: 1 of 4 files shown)


goretkin:
Can you speak to the semantics of this operation? If I got it, this operation is consistent with the definition at https://github.com/JuliaGeometry/Quaternions.jl/blob/master/src/Quaternion.jl#L15 to convert `Complex` to `Quaternion`.

One possible answer to the semantics question is that this multiplication corresponds to rotation in a specific plane. Have there been real cases where you've wanted to do this?

dkarrasch:
The changed methods currently don't work at all, because there is no `Quaternions.imag(::Complex)`. Remember that `imag` is not imported from `Base`, and I think this is done on purpose. I noticed this while writing tests for LinearMaps.jl, but didn't need it in the end, so we can as well close this.

dkarrasch:
Gentle bump. I still needed it in `LinearMaps.jl` tests, in which I needed a non-commutative number type multiplied with complex numbers. I renamed the PR to reflect that this actually fixes broken conversion methods.

dkarrasch:
I guess this is superseded by #52, but I'll leave it open in case there is something controversial about #52.

sethaxen:
Closing this now because #52 was merged.

PR #41, opened 2020-10-05
Add missing Base.imag import

johnhearn:
This adds a missing Base import for `imag()` which also affects promotions from complex numbers.


dkarrasch:
x-ref #29.

PR #42, opened 2020-10-14
Improvements for drawing random numbers

sethaxen:
This PR adds an optional `AbstractRNG` parameter to `quatrand` and `nquatrand`.

It also implements the necessary overloads to call methods like `Base.rand(Quaternion{Float64}, 10)` and `Base.randn(Quaternion{Float64}, 10)`. These behave differently from `quatrand`, instead following the pattern set in base by `Base.rand(<:Complex)`. Namely, `rand` draws each of the four real numbers from the interval `[0,1]`. `randn` draws from the proper Quaternion normal distribution, which requires a scaling by `1/sqrt(4)` (as opposed to `Complex` which uses a scaling of `1/sqrt(2)`).


sethaxen:
@KristofferC can you review this or recommend someone who can?

PR #46, opened 2021-04-15
add matrix exponentials

sw2030:
Quaternion matrix exponential added using a quick detour to complex representation.

### DIFF
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -304,7 +304,7 @@ function _complex_representation(A::Matrix{Quaternion{T}}) where {T}
 	return [a+im*b c+im*d;-c+im*d a-im*b]
 end
 
-function _quaternion_representation(A::Matrix{Complex})
+function _quaternion_representation(A::Matrix{Complex{T}}) where {T}
 	n = round(Int, size(A, 1)/2)
     a = real(A[1:n, 1:n])
     b = imag(A[1:n, 1:n])
@@ -314,4 +314,4 @@ function _quaternion_representation(A::Matrix{Complex})
 end
 
 # A quick way of doing the quaternionic matrix exponential
-exp(A::Matrix{Quaternion}) = _quaternion_representation(exp(_complex_representation(A)))
+exp(A::Matrix{Quaternion{T}}) where {T} = _quaternion_representation(exp(_complex_representation(A)))
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -293,3 +293,25 @@ function slerp(qa::Quaternion{T}, qb::Quaternion{T}, t::T) where {T}
         qa.v3 * ratio_a + qm.v3 * ratio_b,
     )
 end
+
+function _complex_representation(A::Matrix{Quaternion})
+	# convert a quatenion matrix to corresponding complex matrix
+	a = map(t->t.s, A)
+    b = map(t->t.v1, A)
+    c = map(t->t.v2, A)
+    d = map(t->t.v3, A)
+    
+	return [a+im*b c+im*d;-c+im*d a-im*b]
+end
+
+function _quaternion_representation(A::Matrix{Complex})
+	n = round(Int, size(A, 1)/2)
+    a = real(A[1:n, 1:n])
+    b = imag(A[1:n, 1:n])
+    c = real(A[1:n, n+1:2n])
+    d = imag(A[1:n, n+1:2n])
+    return [Quaternion(a[i,j], b[i,j], c[i,j], d[i,j]) for i in 1:n, j in 1:n]
+end
+
+# A quick way of doing the quaternionic matrix exponential
+exp(A::Matrix{Quaternion}) = _quaternion_representation(exp(_complex_representation(A)))
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -310,7 +310,7 @@ function _quaternion_representation(A::Matrix{Complex{T}}) where {T}
     X = @view A[1:n, 1:n]
     Y = @view A[1:n, n+1:2n]
 
-	return [Quaternion(X[i,j].re, X[i,j].im, Y[i,j].re, Y[i,j].im) for i in 1:n, j in 1:n]
+    return [Quaternion(X[i,j].re, X[i,j].im, Y[i,j].re, Y[i,j].im) for i in 1:n, j in 1:n]
 end
 
 # A quick way of doing the quaternionic matrix exponential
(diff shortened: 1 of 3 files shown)


src/Quaternion.jl
```diff
@@ -293,3 +293,25 @@ function slerp(qa::Quaternion{T}, qb::Quaternion{T}, t::T) where {T}
         qa.v3 * ratio_a + qm.v3 * ratio_b,
     )
 end
+
+function _complex_representation(A::Matrix{Quaternion{T}}) where {T}
+	# convert a quatenion matrix to corresponding complex matrix
```
KristofferC, line 298: These two lines have tabs over spaces.

KristofferC:
Would be good with a test as well :)

src/Quaternions.jl
```diff
@@ -4,8 +4,8 @@ module Quaternions
 
   import Base: +, -, *, /, ^
   import Base: abs, abs2, angle, conj, cos, exp, inv, isfinite, log, real, sin, sqrt
-  import Base: convert, promote_rule, float
-  import LinearAlgebra: norm, normalize
+  import Base: convert, promote_rule, float, imag
+  import LinearAlgebra: norm, normalize, exp
```
KristofferC, line 8: `exp` is defined in Base so no need to import it from LinearAlgebra

src/Quaternions.jl
```diff
@@ -4,8 +4,8 @@ module Quaternions
 
   import Base: +, -, *, /, ^
   import Base: abs, abs2, angle, conj, cos, exp, inv, isfinite, log, real, sin, sqrt
-  import Base: convert, promote_rule, float
-  import LinearAlgebra: norm, normalize
+  import Base: convert, promote_rule, float, imag
```
KristofferC, line 7: No need to import `imag`, it is exported from `Base`.

src/Quaternion.jl
```diff
@@ -293,3 +293,25 @@ function slerp(qa::Quaternion{T}, qb::Quaternion{T}, t::T) where {T}
...
+	a = map(t->t.s, A)
+    b = map(t->t.v1, A)
+    c = map(t->t.v2, A)
+    d = map(t->t.v3, A)
+    
+	return [a+im*b c+im*d;-c+im*d a-im*b]
+end
+
+function _quaternion_representation(A::Matrix{Complex{T}}) where {T}
+	n = round(Int, size(A, 1)/2)
```
KristofferC, line 308: tabs

src/Quaternion.jl
```diff
@@ -293,3 +293,25 @@ function slerp(qa::Quaternion{T}, qb::Quaternion{T}, t::T) where {T}
...
+    a = real(A[1:n, 1:n])
```
KristofferC, line 309: Might be better to use a `view` (`@view A[1:n, 1:n]`) and then call `real` / `imag` on the scalar in the `Quaternion` call below.

KristofferC:
It would be good with a test to see that the functionality is working properly and doesn't regress in the future.

src/Quaternions.jl
sw2030: changed to .im instead of imag - seems like there's an issue with importing imag
https://github.com/JuliaGeometry/Quaternions.jl/issues/29

sw2030:
Added testing, fixed spacing and import issues!

KristofferC:
Thanks, it looks like we need to update the CI infrastructure here from Travis to Github Actions. I will do that and then CI should run on this PR.

src/Quaternion.jl
```diff
@@ -293,3 +293,25 @@ function slerp(qa::Quaternion{T}, qb::Quaternion{T}, t::T) where {T}
...
 end
+
+function _complex_representation(A::Matrix{Quaternion{T}}) where {T}
+    # convert a quatenion matrix to corresponding complex matrix
+    a = map(t->t.s, A)
+    b = map(t->t.v1, A)
+    c = map(t->t.v2, A)
+    d = map(t->t.v3, A)
+    
+    return [a+im*b c+im*d;-c+im*d a-im*b]
```
sethaxen, line 304: Currently this allocates a lot of intermediate arrays. What about something like this to allocate just 1?

```suggestion
    n = size(A, 1)
    Ac = Matrix{Complex{T}}(undef, 2n, 2n)
    Ac[1:n, 1:n] .= complex.(getfield.(A, :s), getfield.(A, :v1))
    Ac[1:n, n+1:2n] .= complex.(getfield.(A, :v2), getfield.(A, :v3))
    Ac[n+1:2n, 1:n] .= .- conj.(view(Ac, 1:n, n+1:2n))
    Ac[n+1:2n, n+1:2n] .= conj.(view(Ac, 1:n, 1:n))
    return Ac
```

test/test_Quaternion.jl
```diff
@@ -138,4 +138,13 @@ for _ in 1:100
         @test q ⊗ slerp(q1, q2, t) ≈ slerp(q ⊗ q1, q ⊗ q2, t)
         @test q ⊗ linpol(q1, q2, t) ≈ linpol(q ⊗ q1, q ⊗ q2, t)
     end
+    let # test matrix exponential
```
sethaxen, line 141:
```suggestion
    @testset "exp(::Matrix{Quaternion})" begin  # test matrix exponential
```

test/test_Quaternion.jl
```diff
@@ -138,4 +138,13 @@ for _ in 1:100
...
+        z = zeros(Quaternion{Float64}, 3, 3)
+        id = Matrix{Quaternion{Float64}}(I, 3, 3)
+        d = [Quaternion(randn(4)...) for i in 1:3]
+        expd = exp.(d)
+        D = exp(Array(Diagonal(d)))
```
sethaxen, line 146: Can you include a test for a dense matrix also? Perhaps you could check it using `evalpoly`.

sethaxen (changes requested):
Thanks for the PR! A few suggestions.

mikmoore:
While there could be reasons for a specialized implementation like this (e.g., it looks like this could take advantage of existing complex matrix exponentiation, which is reasonably performant), it appears this PR adds support *exclusively* for `Matrix` with `Quaternion` elements. It would not enable support for other matrix types (e.g., `Hermitian`, `StaticArrays.SMatrix`), so would still leave a lot of holes in the functionality.

This issue could be addressed more generically by instead (or additionally) addressing [JuliaLang #51008](https://github.com/JuliaLang/julia/issues/51008).

PR #52, opened 2021-12-20
improve promotion and equality checking

mikmoore:
This PR is primarily motivated by an issue where `Quaternion(1) == 1` returned `false`. There were two issues at play here.

First, `convert(Quaternion,::Real)` and `convert(Quaternion,::Complex)` were never setting the `.norm` field of the resulting Quaternion. This meant that `Quaternion(1.0).norm == true` while `convert(Quaternion,1.0).norm == false`.

Second, there was no special-cased `==(::Quaternion,::Quaternion)` method. This meant that equality required the `.norm` fields to match.

This PR reroutes `convert` calls through the normal constructors to help `.norm` be set more consistently. It also adds a `==` method for Quaternions so that the `.norm` field is ignored in equality checks. Lastly, it adds some tests along these lines.

### DIFF
--- a/src/DualQuaternion.jl
+++ b/src/DualQuaternion.jl
@@ -18,7 +18,7 @@ DualQuaternion(d1::Dual, d2::Dual, d3::Dual, d4::Dual, n::Bool) =
 
 DualQuaternion(x::Real) = DualQuaternion(Quaternion(x), Quaternion(zero(x)), abs(x) == one(x))
 
-DualQuaternion(d::Dual) = DualQuaternion(d, zero(d), zero(d), zero(d), abs(d) == one(DualNumbers.value(d)))
+DualQuaternion(d::Dual) = DualQuaternion(Quaternion(DualNumbers.value(d)), Quaternion(DualNumbers.epsilon(d)), (DualNumbers.value(d)==one(DualNumbers.value(d))) & iszero(DualNumbers.epsilon(d)))
 
 DualQuaternion(q::Quaternion) = DualQuaternion(q, zero(q), q.norm)
 
--- a/src/DualQuaternion.jl
+++ b/src/DualQuaternion.jl
@@ -21,14 +21,11 @@ DualQuaternion(q::Quaternion) = DualQuaternion(q, zero(q), q.norm)
 
 DualQuaternion(a::Vector) = DualQuaternion(zero(Quaternion{typeof(a[1])}), Quaternion(a))
 
-convert(::Type{DualQuaternion{T}}, x::Real) where {T} =
-    DualQuaternion(convert(Quaternion{T}, x), convert(Quaternion{T}, 0))
+convert(::Type{DualQuaternion{T}}, x::Real) where {T} = DualQuaternion(convert(T, x))
 
-convert(::Type{DualQuaternion{T}}, d::Dual) where {T} =
-    DualQuaternion(convert(Dual{T}, d), convert(Dual{T}, 0), convert(Dual{T}, 0), convert(Dual{T}, 0))
+convert(::Type{DualQuaternion{T}}, d::Dual) where {T} = DualQuaternion(convert(Dual{T}, d))
 
-convert(::Type{DualQuaternion{T}}, q::Quaternion) where {T} =
-    DualQuaternion(convert(Quaternion{T}, q), convert(Quaternion{T}, 0), q.norm)
+convert(::Type{DualQuaternion{T}}, q::Quaternion) where {T} = DualQuaternion(convert(Quaternion{T}, q))
 
 convert(::Type{DualQuaternion{T}}, q::DualQuaternion{T}) where {T <: Real} = q
 
@@ -112,6 +109,7 @@ end
                                                               dq.q0 * dw.qe + dq.qe * dw.q0,
                                                               dq.norm && dw.norm)
 (/)(dq::DualQuaternion, dw::DualQuaternion) = dq * inv(dw)
+(==)(q::DualQuaternion, w::DualQuaternion) = (q.q0 == w.q0) & (q.qe == w.qe) # ignore .norm field
 
 function angleaxis(dq::DualQuaternion)
   tq = dq.qe * conj(dq.q0)
--- a/src/DualQuaternion.jl
+++ b/src/DualQuaternion.jl
@@ -6,16 +6,19 @@ struct DualQuaternion{T<:Real} <: Number
   norm::Bool
 end
 
-DualQuaternion(q0::Quaternion, qe::Quaternion, n::Bool = false) =
+DualQuaternion(q0::Quaternion, qe::Quaternion, n::Bool = q0.norm) =
   DualQuaternion(promote(q0, qe)..., n)
 
-DualQuaternion(d1::Dual, d2::Dual, d3::Dual, d4::Dual, n::Bool = false) =
-  DualQuaternion(Quaternion(d1.re, d2.re, d3.re, d4.re, n),
-                  Quaternion(d1.du, d2.du, d3.du, d4.du), n)
+DualQuaternion(d1::Dual, d2::Dual, d3::Dual, d4::Dual) =
+  DualQuaternion(Quaternion(DualNumbers.value(d1), DualNumbers.value(d2), DualNumbers.value(d3), DualNumbers.value(d4)),
+                  Quaternion(DualNumbers.epsilon(d1), DualNumbers.epsilon(d2), DualNumbers.epsilon(d3), DualNumbers.epsilon(d4)))
+DualQuaternion(d1::Dual, d2::Dual, d3::Dual, d4::Dual, n::Bool) =
+  DualQuaternion(Quaternion(DualNumbers.value(d1), DualNumbers.value(d2), DualNumbers.value(d3), DualNumbers.value(d4), n),
+                  Quaternion(DualNumbers.epsilon(d1), DualNumbers.epsilon(d2), DualNumbers.epsilon(d3), DualNumbers.epsilon(d4)), n)
 
 DualQuaternion(x::Real) = DualQuaternion(Quaternion(x), Quaternion(zero(x)), abs(x) == one(x))
 
-DualQuaternion(d::Dual) = DualQuaternion(d, zero(d), zero(d), zero(d), abs(d) == one(d.re))
+DualQuaternion(d::Dual) = DualQuaternion(d, zero(d), zero(d), zero(d), abs(d) == one(d))
 
 DualQuaternion(q::Quaternion) = DualQuaternion(q, zero(q), q.norm)
 
--- a/src/DualQuaternion.jl
+++ b/src/DualQuaternion.jl
@@ -38,9 +38,7 @@ promote_rule(::Type{DualQuaternion{T}}, ::Type{S}) where {T <: Real, S <: Real} 
 promote_rule(::Type{Quaternion{T}}, ::Type{DualQuaternion{S}}) where {T <: Real, S <: Real} = DualQuaternion{promote_type(T, S)}
 promote_rule(::Type{DualQuaternion{T}}, ::Type{DualQuaternion{S}}) where {T <: Real, S <: Real} = DualQuaternion{promote_type(T, S)}
 
-dualquat(q1, q2, n=false) = DualQuaternion(q1, q2, n)
-dualquat(d1, d2, d3, d4, n=false) = DualQuaternion(d1, d2, d3, d4, n)
-dualquat(x) = DualQuaternion(x)
+const dualquat = DualQuaternion
 
 function show(io::IO, dq::DualQuaternion)
   show(io, dq.q0)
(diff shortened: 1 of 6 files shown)


src/Quaternion.jl
```diff
@@ -16,10 +16,8 @@ Quaternion(z::Complex) = Quaternion(z.re, z.im, zero(z.re), zero(z.re), abs(z) =
 Quaternion(s::Real, a::Vector) = Quaternion(s, a[1], a[2], a[3])
 Quaternion(a::Vector) = Quaternion(0, a[1], a[2], a[3])
 
-convert(::Type{Quaternion{T}}, x::Real) where {T} =
-    Quaternion(convert(T, x), convert(T, 0), convert(T, 0), convert(T, 0))
-convert(::Type{Quaternion{T}}, z::Complex) where {T} =
-    Quaternion(convert(T, real(z)), convert(T, imag(z)), convert(T, 0), convert(T, 0))
+convert(::Type{Quaternion{T}}, x::Real) where {T} = Quaternion(convert(T,x))
+convert(::Type{Quaternion{T}}, z::Complex) where {T} = Quaternion(convert(Complex{T},z))
```
dkarrasch, line 20:
```suggestion
convert(::Type{Quaternion{T}}, x::Real) where {T} = Quaternion(convert(T, x))
convert(::Type{Quaternion{T}}, z::Complex) where {T} = Quaternion(convert(Complex{T}, z))
```

src/Quaternion.jl
```diff
@@ -96,6 +94,8 @@ end
                                                q.norm && w.norm)
 (/)(q::Quaternion, w::Quaternion) = q * inv(w)
 
+(==)(q::Quaternion{T}, w::Quaternion{T}) where T = (q.s == w.s && q.v1 == w.v1 && q.v2 == w.v2 && q.v3 == w.v3) # ignore .norm field
```
dkarrasch, line 97:
```suggestion
(==)(q::Quaternion, w::Quaternion) = (q.s == w.s && q.v1 == w.v1 && q.v2 == w.v2 && q.v3 == w.v3) # ignore .norm field
```

dkarrasch (commented):
Just a few minor style comments.

dkarrasch:
This should close #40, IIUC.

mikmoore:
I noticed that `&` is probably better than `&&` for short chains of simple comparisons (no branches), so I changed that slightly although it can certainly be reverted.
I also extended this PR to include the matching changes to `Octonion` and `DualQuaternion`, which I had neglected before.

src/DualQuaternion.jl
```diff
@@ -112,6 +109,7 @@ end
                                                               dq.q0 * dw.qe + dq.qe * dw.q0,
                                                               dq.norm && dw.norm)
 (/)(dq::DualQuaternion, dw::DualQuaternion) = dq * inv(dw)
+(==)(q::DualQuaternion, w::DualQuaternion) = (q.q0 == w.q0) & (q.qe == w.qe) # ignore .norm field
```
sethaxen, line 112: Ah, I see you mentioned `&` is better in https://github.com/JuliaGeometry/Quaternions.jl/pull/52#issuecomment-1003257187. Can you explain the reasoning?

src/Octonion.jl
```diff
@@ -113,6 +110,9 @@ end
 
 (/)(o::Octonion, w::Octonion) = o * inv(w)
 
+(==)(q::Octonion, w::Octonion) = (q.s == w.s)   & (q.v1 == w.v1) & (q.v2 == w.v2) & (q.v3 == w.v3) &
```
sethaxen, line 113:
```suggestion
(==)(q::Octonion, w::Octonion) = (q.s == w.s) & (q.v1 == w.v1) & (q.v2 == w.v2) & (q.v3 == w.v3) &
```

sethaxen (changes requested):
Thanks for the PR! Can you add tests for `==` for `Octonion` and `DualQuaternion` as well? Also, most of the conversions modified here are not yet covered by tests, and they should be tested as well.

src/DualQuaternion.jl
mikmoore: The `code_native` for the `Quaternion` `&` version is great. It compare all four fields via a single SIMD instruction and has no branches. Meanwhile, the same function using `&&` requires a check-then-branch for each field because it must short-circuit the later comparisons on inequality.

The benchmarking isn't that different between the two. About 1.9ns for `&` and 2.1ns for `&&`. The benchmark I used would have let the branch predictor learn the `&&` version, so `&&` might perform worse in the wild. Also, `&` is easier to inline since it's shorter.

The same story holds for `DualQuaternion`. I didn't check `Octonion` but the story should be similar to `DualQuaternion` or slightly better (since the SIMD loads don't have to hopscotch over the `norm` field of the first Quaternion.)

sethaxen: Thanks!

sethaxen (approved):

sethaxen (changes requested):
Sorry, accidentally approved before the missing tests were added.

test/test_Quaternion.jl
```diff
@@ -24,6 +24,25 @@ for _ in 1:10, T in (Float32, Float64, Int32, Int64)
     test_multiplicative(c1, c2, +, Quaternion)
 end
 
+let # test promotions and equalities
```
sethaxen, line 27: It's more useful to declare a testset:
```suggestion
@testset "promotions and equalities" begin
```

sethaxen: Also, can you create a testset for testing the conversions directly instead of just through promotions? The current tests here miss the following conversions, and this way we guarantee to never miss a conversion in the tests.
```julia
convert(::Type{DualQuaternion{T}}, d::Dual) where {T}
convert(::Type{Octonion{T}}, z::Complex) where {T}
```

sethaxen (changes requested):

mikmoore:
I noted that the internal(?) shorthand functions `quat`, `dualquat`, and `octo` had the same issue of defaulting the `norm` field to false, in addition to being totally redundant. I redefined them to be constants equal to their corresponding type, which should fix that. In the grander scheme I think they really had ought to be removed, but that's beyond the scope of this PR.

I added explicit conversion tests. They revealed that `DualQuaternion` was broken when given `DualNumbers.Dual` because the field names were wrong. I believe I fixed that as well, but somebody should check because dual numbers aren't really my thing.

src/DualQuaternion.jl
```diff
@@ -6,29 +6,29 @@ struct DualQuaternion{T<:Real} <: Number
...
+  DualQuaternion(Quaternion(DualNumbers.value(d1), DualNumbers.value(d2), DualNumbers.value(d3), DualNumbers.value(d4)),
+                  Quaternion(DualNumbers.epsilon(d1), DualNumbers.epsilon(d2), DualNumbers.epsilon(d3), DualNumbers.epsilon(d4)))
+DualQuaternion(d1::Dual, d2::Dual, d3::Dual, d4::Dual, n::Bool) =
+  DualQuaternion(Quaternion(DualNumbers.value(d1), DualNumbers.value(d2), DualNumbers.value(d3), DualNumbers.value(d4), n),
+                  Quaternion(DualNumbers.epsilon(d1), DualNumbers.epsilon(d2), DualNumbers.epsilon(d3), DualNumbers.epsilon(d4)), n)
 
 DualQuaternion(x::Real) = DualQuaternion(Quaternion(x), Quaternion(zero(x)), abs(x) == one(x))
 
-DualQuaternion(d::Dual) = DualQuaternion(d, zero(d), zero(d), zero(d), abs(d) == one(d.re))
+DualQuaternion(d::Dual) = DualQuaternion(d, zero(d), zero(d), zero(d), abs(d) == one(d))
```
sethaxen, line 21:
```suggestion
DualQuaternion(d::Dual) = DualQuaternion(d, zero(d), zero(d), zero(d), abs(d) == one(DualNumbers.value(d)))
```

src/DualQuaternion.jl
```diff
@@ -6,29 +6,29 @@ struct DualQuaternion{T<:Real} <: Number
   norm::Bool
 end
 
-DualQuaternion(q0::Quaternion, qe::Quaternion, n::Bool = false) =
+DualQuaternion(q0::Quaternion, qe::Quaternion, n::Bool = q0.norm) =
```
sethaxen, line 9: This does not seem to be consistent with how `norm` is defined for `DualQuaternion`. `norm` returns a `Dual` number, and the `DualQuaternion` is normalized when the real part of the norm is 1 and the infinitesimal part is 0.

sethaxen: I think we only want to set `norm=true` when its value can be determined from either the types being provided or from the value of other `norm` fields, which I don't think we can do here.

sethaxen (commented):
Aliasing the shorthand functions would be a breaking change. While the public API of this package is not documented, they are exported functions and should be considered part of the API. As for their use `quat` is to `Quaternion` as `complex` is to `Complex`. The former is a method the user can overload to quaternion-ify anything, while the latter is a constructor that should return a `Quaternion` (and probably should not be overloaded). You could open an issue to discuss removing them, but that really should not be in this PR.

sethaxen:
Also, can you rebase on master and resolve merge conflicts?

src/DualQuaternion.jl
mikmoore: I have zero knowledge of dual quaternions, hence my mistake. It appears this has veered wildly beyond the scope of both this PR and my understanding and I'll revert my change.

Off-topic discussion: wouldn't your statement imply that `n::Bool = q0.norm & iszero(qe)` would be the proper initialization? More importantly, doesn't any requirement on the infinitesimal part imply that `normalize(::DualQuaternion)` and friends are incorrect in setting `norm=true` without examining the infinitesimal? According to [https://en.wikipedia.org/wiki/Dual_quaternion#Norm](url), however, it appears to be slightly less restrictive than what you have claimed. It says that a dual quaternion `A+eB` is unit when `(AA*,AB*+BA*)=(1,0)`.

mikmoore:
> As for their use `quat` is to `Quaternion` as `complex` is to `Complex`.

I see. In that case, I changed them to not default `norm=false` but instead to only forward a norm flag if provided one. This will let them inherit default `norm` behavior from the relevant constructors. Without this change, for example, `Quaternion(1)` and `quat(1)` resulted in different `norm` flags.

src/DualQuaternion.jl
sethaxen: TBH, I don't have much experience with dual quaternions either, so I can't really answer the questions without digging into the code and the theory a bit more. My point was just that the change would create inconsistencies in the code base.

test/test_Quaternion.jl
```diff
@@ -1,5 +1,6 @@
 
 using Quaternions: argq
+import DualNumbers
```
sethaxen, line 3: `import` should be avoided unless overloading

```suggestion
using DualNumbers
```

sethaxen (commented):
If you browse through the code, CodeCov has annotated a few lines changed here that are not covered by the tests. Can you check that they are? Also, can you add tests that the shorthand cases you mentioned where `norm` was set differently than by the constructor are also tested?

test/test_Quaternion.jl
```diff
@@ -37,7 +38,69 @@ for _ in 1:10, T in (Float32, Float64, Int32, Int64)
...
+    @test DualQuaternion(Quaternion(1,2,3,4),Quaternion(5,6,7,8)) == DualQuaternion(Quaternion(1.0,2,3,4),Quaternion(5,6,7,8))
+    @test DualQuaternion(Quaternion(1,2,3,4),Quaternion(5,6,7,8)) != DualQuaternion(Quaternion(1.0,2,3,4),Quaternion(1,2,3,4))
+    @test dualquat(Quaternion(1,0,0,0)) == Quaternion(1,0,0,0)
+    @test dualquat(Quaternion(1,2,3,4)) == Quaternion(1,2,3,4)
+    @test dualquat(Quaternion(1,0,0,0)) === DualQuaternion(Quaternion(1,0,0,0)) # checking the .norm field in particular
+    @test dualquat(Quaternion(1,2,3,4)) === DualQuaternion(Quaternion(1,2,3,4))
+    @test dualquat(1) === DualQuaternion(1)
+    @test dualquat(Dual(1,2)) === DualQuaternion(Dual(1,2))
+    @test dualquat(Dual(1,2),Dual(0),Dual(0),Dual(0)) === DualQuaternion(Dual(1,2),Dual(0),Dual(0),Dual(0))
+    @test dualquat(Quaternion(1,0,0,0),Quaternion(0),false).norm == false # respect the .norm input (even if wrong)
```
sethaxen, line 66:
```suggestion
    @test dualquat(Quaternion(1,2,3,4),Quaternion(5,6,7,8)) == DualQuaternion(Quaternion(1,2,3,4),Quaternion(5,6,7,8))
    @test dualquat(Quaternion(1,0,0,0),Quaternion(0)).norm == false
    @test dualquat(Quaternion(1,0,0,0),Quaternion(0),false).norm == false # respect the .norm input (even if wrong)
```

sethaxen (approved):
LGTM, thanks!

PR #53, opened 2021-12-21
fix piracy and type instability

mikmoore:
This PR is intended to remove the piracy of `LinearAlgebra.normalize(::Vector)` that was previously being committed by Quaternions. This piracy would make `normalize(zeros(3))` return a vector of `0.0` when it would normally return a vector of `NaN`.

The guilty method was only used in `qrotation(::Vector,::Any)` so it was easy enough to remove it and provide the matching calculation in its place. I also noticed that `qrotation([0,0,0],1.0) == Quaternion(0.8775825618903728, 0.0, 0.0, 0.0, true)`, which is wrong in multiple capacities. I have made it an error to use this method with a zero vector in the first argument. A potential alternative would be to force the rotation angle to zero in this case, but I opted for the error since a zero axis seemed more likely to be used in error than deliberately (given that `qrotation(::Vector)` is available for the latter case).

Along the way, I noticed that the related method `qrotation(::Vector)` was type-unstable for zero inputs and that input vectors with `NaN` entries failed to propagate their `NaN`s. Both of these issues are resolved by making it mirror the changes already made to the 2-arg method.

### DIFF
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -209,11 +209,12 @@ function qrotation(axis::Vector{T}, theta) where {T <: Real}
     end
     normaxis = norm(axis)
     if iszero(normaxis)
-        error("must provide nonzero rotation axis")
+        normaxis = one(normaxis)
+        theta = zero(theta)
     end
     s,c = sincos(theta / 2)
-    scaleby = s / normaxis # normaxis not zero by earlier check
-    Quaternion(c * one(scaleby), scaleby * axis[1], scaleby * axis[2], scaleby * axis[3], true)
+    scaleby = s / normaxis
+    Quaternion(c, scaleby * axis[1], scaleby * axis[2], scaleby * axis[3], true)
 end
 
 # Variant of the above where norm(rotvec) encodes theta
@@ -224,7 +225,7 @@ function qrotation(rotvec::Vector{T}) where {T <: Real}
     theta = norm(rotvec)
     s,c = sincos(theta / 2)
     scaleby = s / (iszero(theta) ? one(theta) : theta)
-    return Quaternion(c, scaleby * rotvec[1], scaleby * rotvec[2], scaleby * rotvec[3], true)
+    Quaternion(c, scaleby * rotvec[1], scaleby * rotvec[2], scaleby * rotvec[3], true)
 end
 
 function qrotation(dcm::Matrix{T}) where {T<:Real}
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -193,10 +193,13 @@ function qrotation(axis::Vector{T}, theta) where {T <: Real}
     if length(axis) != 3
         error("Must be a 3-vector")
     end
-    u = normalize(axis)
-    thetaT = convert(eltype(u), theta)
-    s = sin(thetaT / 2)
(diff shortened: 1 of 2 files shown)


src/Quaternion.jl
```diff
@@ -193,10 +193,13 @@ function qrotation(axis::Vector{T}, theta) where {T <: Real}
...
         error("Must be a 3-vector")
     end
-    u = normalize(axis)
-    thetaT = convert(eltype(u), theta)
-    s = sin(thetaT / 2)
-    Quaternion(cos(thetaT / 2), s * u[1], s * u[2], s * u[3], true)
+    normaxis = norm(axis)
+    if iszero(normaxis)
+        error("must provide nonzero rotation axis")
+    end
```
sethaxen, line 199: Why this change in behavior? Currently if one provides a zero axis, i.e. corresponding to no rotation at all, `qrotation` returns `Quaternion(1)`, the multiplicative identity, which seems correct.

sethaxen (commented):
Thanks for the PR! One question about the change in behavior of `qrotation`.

src/Quaternion.jl
sethaxen: I'd prefer you open an issue to discuss it. This would be a breaking change (the current behavior is not a bug), and I don't want to hold up the main fix from this PR being merged.

src/Quaternion.jl
```diff
@@ -207,9 +207,13 @@ function qrotation(axis::Vector{T}, theta) where {T <: Real}
...
-    u = normalize(axis)
-    s = sin(theta / 2)
-    Quaternion(cos(theta / 2), s * u[1], s * u[2], s * u[3], true)
+    normaxis = norm(axis)
+    if iszero(normaxis)
+        error("must provide nonzero rotation axis")
+    end
+    s,c = sincos(theta / 2)
+    scaleby = s / normaxis # normaxis not zero by earlier check
+    Quaternion(c, scaleby * axis[1], scaleby * axis[2], scaleby * axis[3], true)
```
sethaxen, line 216:
```suggestion
    Quaternion(c * one(scaleby), scaleby * axis[1], scaleby * axis[2], scaleby * axis[3], true)
```

src/Quaternion.jl
mikmoore: I'll leave it alone. It would indeed be disruptive to throw an error where one was not thrown previously. As suggested, I've changed the computation to instead force the rotation to zero when the axis is zero.

src/Quaternion.jl
```diff
@@ -207,9 +207,14 @@ function qrotation(axis::Vector{T}, theta) where {T <: Real}
     if length(axis) != 3
         error("Must be a 3-vector")
     end
-    u = normalize(axis)
-    s = sin(theta / 2)
-    Quaternion(cos(theta / 2), s * u[1], s * u[2], s * u[3], true)
+    normaxis = norm(axis)
+    if iszero(normaxis)
+        normaxis = one(normaxis)
```
sethaxen, line 212: This ensures type-stability if the eltype of `axis` is a dimensional number
```suggestion
        normaxis = oneunit(normaxis)
```

sethaxen (changes requested):

sethaxen (approved):

PR #56, opened 2022-02-19
Extend all analytic functions in base for quaternions

### LINKED ISSUE #39: in exp() function, eps yields an error when used on an integer
Using the quaternion package and instantiating quaternions of the shape
a = Quaternion(1,1,1,1)
the command
exp(a) 
returned an error of the function eps : "no method matching eps(::Int64)". Maybe an automatic convert of the coordinates of the quaternion to Float could solve the problem? 
In the meantime, I just changed my initialization for:
a = Quaternion(1.0,1.0,1.0,1.0) which lifted the error.

### LINKED ISSUE #55: Extend all analytic functions in base for quaternions
All complex analytic functions `f` in Base can be extended to the quaternions with the following function:

```julia
using Quaternions

function quat_analytic(f, q::Quaternion)
    a = Quaternions.abs_imag(q)
    z = complex(q.s, a)
    fz = f(z)
    wr, wi = reim(fz)
    scale = wi / a
    if a > 0
        return Quaternion(wr, scale * q.v1, scale * q.v2, scale * q.v3)
    else  # quaternion may be real or complex
        return Quaternion(wr, oftype(scale, wi), zero(scale), zero(scale))
    end
end
```

The function is adapted from Theorem 5 of [doi.org/10.1017/S0305004100055638](https://doi.org/10.1017/S0305004100055638). It can also be derived a completely different way by mapping a quaternion to a 2x2 skew-Hermitian matrix whose product preserves the Hamilton product, extending the analytic function to a matrix function, and then mapping back to the equivalent quaternions.

This function gives equivalent results to the current functions implementations here, but it's faster than all but `log`:

```julia
julia> using BenchmarkTools

julia> q = quatrand()
Quaternion{Float64}(-1.76527498999994, -0.9985320755133996, -0.606774830599748, 1.0149711723420227, false)

julia> @btime exp($q)
  31.384 ns (0 allocations: 0 bytes)
Quaternion{Float64}(0.003950512836446494, -0.11038429890473422, -0.0670768780605819, 0.11220158472113788, false)

julia> @btime $quat_analytic(exp, $q)
  28.991 ns (0 allocations: 0 bytes)
Quaternion{Float64}(0.003950512836446494, -0.11038429890473421, -0.06707687806058188, 0.11220158472113786, false)

julia> @btime log($q)
  74.247 ns (0 allocations: 0 bytes)
Quaternion{Float64}(0.8534278686043146, -1.5624466956559513, -0.9494470456448084, 1.5881696675557215, false)

julia> @btime $quat_analytic(log, $q)
  81.745 ns (0 allocations: 0 bytes)
Quaternion{Float64}(0.8534278686043146, -1.5624466956559513, -0.9494470456448083, 1.5881696675557215, false)

julia> @btime sqrt($q)
  131.847 ns (0 allocations: 0 bytes)
Quaternion{Float64}(0.539632108658829, -0.9251970550780371, -0.5622115704973448, 0.9404288181299799, false)

julia> @btime $quat_analytic(sqrt, $q)
  30.934 ns (0 allocations: 0 bytes)
Quaternion{Float64}(0.5396321086588289, -0.9251970550780371, -0.5622115704973447, 0.9404288181299798, false)

julia> @btime sin($q)
  158.434 ns (0 allocations: 0 bytes)
Quaternion{Float64}(-2.410402900676647, 0.2797836189575866, 0.1700152275131357, -0.28438977044324726, false)

julia> @btime $quat_analytic(sin, $q)
  48.684 ns (0 allocations: 0 bytes)
Quaternion{Float64}(-2.410402900676648, 0.2797836189575863, 0.17001522751313533, -0.2843897704432474, false)

julia> quat_analytic(tan, q) ≈ sin(q) / cos(q) ≈ cos(q) \ sin(q)
true

julia> quat_analytic(sin, quat_analytic(asin, q)) ≈ q
true

julia> quat_analytic(cosh, quat_analytic(acosh, q)) ≈ q
true

julia> quat_analytic(exp10, q) ≈ 10^q
true

julia> quat_analytic(inv, q) ≈ inv(q)
true
```

sethaxen:
Fixes #55. Also fixes #39 as a bonus.

### DIFF
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -19,7 +19,7 @@ Quaternion(a::Vector) = Quaternion(0, a[1], a[2], a[3])
 convert(::Type{Quaternion{T}}, x::Real) where {T} =
     Quaternion(convert(T, x), convert(T, 0), convert(T, 0), convert(T, 0))
 convert(::Type{Quaternion{T}}, z::Complex) where {T} =
-    Quaternion(convert(T, real(z)), convert(T, imag(z)), convert(T, 0), convert(T, 0))
+    Quaternion(convert(T, real(z)), convert(T, Base.imag(z)), convert(T, 0), convert(T, 0))
 convert(::Type{Quaternion{T}}, q::Quaternion{T}) where {T <: Real} = q
 convert(::Type{Quaternion{T}}, q::Quaternion) where {T} =
     Quaternion(convert(T, q.s), convert(T, q.v1), convert(T, q.v2), convert(T, q.v3), q.norm)
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -183,8 +183,8 @@ for f in (@static(VERSION ≥ v"1.6" ? (:sincos, :sincospi) : (:sincos,)))
                 )
             else
                 return (
-                    Quaternion(sr, oftype(sscale, si), zero(scale), zero(scale)),
-                    Quaternion(cr, oftype(cscale, ci), zero(scale), zero(scale)),
+                    Quaternion(sr, oftype(sscale, si), zero(sscale), zero(sscale)),
+                    Quaternion(cr, oftype(cscale, ci), zero(cscale), zero(cscale)),
                 )
             end
         end
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -199,8 +199,6 @@ function log(q::Quaternion)
     return Quaternion(log(a), q.v1 * scale, q.v2 * scale, q.v3 * scale)
 end
 
-(^)(q::Quaternion{T}, p::T) where {T<:Real} = extend_analytic(Base.Fix2(^, p), q)
-(^)(q::Quaternion{T}, p::AbstractFloat) where {T<:Real} = extend_analytic(Base.Fix2(^, p), q)
 (^)(q::Quaternion, w::Quaternion) = exp(w * log(q))
 
 function linpol(p::Quaternion, q::Quaternion, t::Real)
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -154,7 +154,7 @@ if VERSION ≥ v"1.6"
     !!! Note
         This is not equivalent to `exp(π*im*q)`. See [cis(::Quaternion)](@ref) for details.
     """
-    cispi(q::Quaternion) = extend_analytic(cispi, q)
+    Base.cispi(q::Quaternion) = extend_analytic(cispi, q)
 end
 
 for f in (
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -130,8 +130,35 @@ end
 _isexpfun(::Union{typeof(exp),typeof(exp2),typeof(exp10)}) = true
 _isexpfun(::Any) = false
 
+"""
+    cis(q::Quaternion)
+
+Return ``\\exp(u * q)``, where ``u`` is the normalized imaginary part of `q`.
+
+Let ``q = s + a u``, where ``s`` is the real part, ``u`` is a pure unit quaternion,
+and ``a \\ge 0`` is the magnitude of the imaginary part of ``q``.
+
+!!! Note
+    This is the extension of `cis(z)` for complex `z` to the quaternions and is not
+    equivalent to `exp(im * q)`. As a result, `cis(Quaternion(z)) ≠ cis(z)` when
+    `imag(z) < 0`.
+"""
+cis(q::Quaternion)
+
+if VERSION ≥ v"1.6"
+    """
+        cispi(q::Quaternion)
+
+    Compute `cis(π * q)` more accurately.
+
+    !!! Note
+        This is not equivalent to `exp(π*im*q)`. See [cis(::Quaternion)](@ref) for details.
+    """
+    cispi(q::Quaternion) = extend_analytic(cispi, q)
+end
+
 for f in (
-    :sqrt, :exp, :exp2, :exp10, :expm1, :log2, :log10, :log1p,
+    :sqrt, :exp, :exp2, :exp10, :expm1, :log2, :log10, :log1p, :cis,
     :sin, :cos, :tan, :asin, :acos, :atan, :sinh, :cosh, :tanh, :asinh, :acosh, :atanh,
     :csc, :sec, :cot, :acsc, :asec, :acot, :csch, :sech, :coth, :acsch, :asech, :acoth,
     :sinpi, :cospi,
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -166,15 +166,10 @@ end
 
 function log(q::Quaternion)
     q, a = normalizea(q)
-    s = q.s
     M = abs_imag(q)
-    th = atan(M, s)
-    if M > 0
-        M = th / M
-        return Quaternion(log(a), q.v1 * M, q.v2 * M, q.v3 * M)
-    else
-        return Quaternion(log(a), th, 0.0, 0.0)
-    end
+    theta = atan(M, q.s)
+    scale = theta / ifelse(iszero(M), oneunit(M), M)
+    return Quaternion(log(a), q.v1 * scale, q.v2 * scale, q.v3 * scale)
 end
 
 (^)(q::Quaternion{T}, p::T) where {T<:Real} = extend_analytic(Base.Fix2(^, p), q)
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -133,6 +133,7 @@ _isexpfun(::Any) = false
 for f in (
     :sqrt, :exp, :exp2, :exp10, :expm1, :log2, :log10, :log1p,
     :sin, :cos, :tan, :asin, :acos, :atan, :sinh, :cosh, :tanh, :asinh, :acosh, :atanh,
+    :csc, :sec, :cot, :acsc, :asec, :acot, :csch, :sech, :coth, :acsch, :asech, :acoth,
     :sinpi, :cospi,
 )
     @eval Base.$f(q::Quaternion) = extend_analytic($f, q)
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -139,7 +139,7 @@ for f in (
     @eval Base.$f(q::Quaternion) = extend_analytic($f, q)
 end
 
-for f in (:sincos, :sincospi)
+for f in (@static(VERSION ≥ v"1.6" ? (:sincos, :sincospi) : (:sincos,)))
     @eval begin
         function Base.$f(q::Quaternion)
             a = abs_imag(q)
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -110,6 +110,54 @@ end
 
 argq(q::Quaternion) = normalizeq(Quaternion(0, q.v1, q.v2, q.v3))
 
+# extend complex analytic function f to the quaternion q
+# adapted from Theorem 5 of doi.org/10.1017/S0305004100055638
+function extend_analytic(f, q::Quaternion)
+    a = abs_imag(q)
+    z = complex(q.s, a)
+    w = f(z)
+    wr, wi = reim(w)
+    scale = wi / a
+    if a > 0
+        return Quaternion(wr, scale * q.v1, scale * q.v2, scale * q.v3)
+    else  # quaternion may be real or complex
+        return Quaternion(wr, oftype(scale, wi), zero(scale), zero(scale))
+    end
+end
+
+for f in (
+    :exp2, :exp10, :expm1, :log2, :log10, :log1p,
+    :tan, :asin, :acos, :atan, :sinh, :cosh, :tanh, :asinh, :acosh, :atanh,
+    :sinpi, :cospi,
+)
+    @eval Base.$f(q::Quaternion) = extend_analytic($f, q)
+end
+
+for f in (:sincos, :sincospi)
+    @eval begin
+        function Base.$f(q::Quaternion)
+            a = abs_imag(q)
+            z = complex(q.s, a)
+            s, c = $f(z)
+            sr, si = reim(s)
+            cr, ci = reim(c)
+            sscale = si / a
+            cscale = ci / a
+            if a > 0
+                return (
+                    Quaternion(sr, sscale * q.v1, sscale * q.v2, sscale * q.v3),
+                    Quaternion(cr, cscale * q.v1, cscale * q.v2, cscale * q.v3),
+                )
+            else
+                return (
+                    Quaternion(sr, oftype(sscale, si), zero(scale), zero(scale)),
+                    Quaternion(cr, oftype(cscale, ci), zero(scale), zero(scale)),
+                )
+            end
+        end
+    end
+end
+
 function exp(q::Quaternion)
     s = q.s
     se = exp(s)
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -114,20 +114,25 @@ argq(q::Quaternion) = normalizeq(Quaternion(0, q.v1, q.v2, q.v3))
 # adapted from Theorem 5 of doi.org/10.1017/S0305004100055638
 function extend_analytic(f, q::Quaternion)
     a = abs_imag(q)
-    z = complex(q.s, a)
+    s = q.s
+    z = complex(s, a)
     w = f(z)
     wr, wi = reim(w)
     scale = wi / a
+    norm = _isexpfun(f) && abs(s) < eps(float(typeof(s)))
     if a > 0
-        return Quaternion(wr, scale * q.v1, scale * q.v2, scale * q.v3)
+        return Quaternion(wr, scale * q.v1, scale * q.v2, scale * q.v3, norm)
     else  # quaternion may be real or complex
-        return Quaternion(wr, oftype(scale, wi), zero(scale), zero(scale))
+        return Quaternion(wr, oftype(scale, wi), zero(scale), zero(scale), norm)
     end
 end
 
+_isexpfun(::Union{typeof(exp),typeof(exp2),typeof(exp10)}) = true
+_isexpfun(::Any) = false
+
 for f in (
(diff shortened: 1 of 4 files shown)


src/Quaternion.jl
```diff
@@ -110,15 +110,58 @@ end
...
+for f in (
+    :sqrt, :exp, :exp2, :exp10, :expm1, :log2, :log10, :log1p,
+    :sin, :cos, :tan, :asin, :acos, :atan, :sinh, :cosh, :tanh, :asinh, :acosh, :atanh,
+    :csc, :sec, :cot, :acsc, :asec, :acot, :csch, :sech, :coth, :acsch, :asech, :acoth,
+    :sinpi, :cospi,
+)
+    @eval Base.$f(q::Quaternion) = extend_analytic($f, q)
+end
+
+for f in (@static(VERSION ≥ v"1.6" ? (:sincos, :sincospi) : (:sincos,)))
```
hyrodium, line 142: I think we can drop support under v1.6 because we have v1.6 LTS. (`Project.toml` will also needed to be updated)
```suggestion
for f in (:sincos, :sincospi)
```

src/Quaternion.jl
```diff
@@ -19,7 +19,7 @@ Quaternion(a::Vector) = Quaternion(0, a[1], a[2], a[3])
 convert(::Type{Quaternion{T}}, x::Real) where {T} =
     Quaternion(convert(T, x), convert(T, 0), convert(T, 0), convert(T, 0))
 convert(::Type{Quaternion{T}}, z::Complex) where {T} =
-    Quaternion(convert(T, real(z)), convert(T, imag(z)), convert(T, 0), convert(T, 0))
+    Quaternion(convert(T, real(z)), convert(T, Base.imag(z)), convert(T, 0), convert(T, 0))
```
hyrodium, line 22: Does this resolves #29 and #62?

test/test_Quaternion.jl
```diff
@@ -67,22 +67,79 @@ let # test rotations
...
+    @test q^1.3 ≈ exp(1.3 * log(q))
+    @test q^7.8 ≈ exp(7.8 * log(q))
+    @test q^1.3f0 ≈ exp(1.3f0 * log(q))
+    @test q^7.8f0 ≈ exp(7.8f0 * log(q))
+end
 
+let
+    # https://github.com/JuliaGeometry/Quaternions.jl/issues/39
+    @test exp(Quaternion(1,1,1,1)) ≈ exp(Quaternion(1.0,1.0,1.0,1.0))
+end
```
hyrodium, line 140:
```suggestion
@testset "exp(::Quaternion{Int})" begin
    # https://github.com/JuliaGeometry/Quaternions.jl/issues/39
    @test exp(Quaternion(1,1,1,1)) ≈ exp(Quaternion(1.0,1.0,1.0,1.0))
end
```
Do we need this `let` block here? I thought it's better to remove `let` or replace it with `@testset`.

test/test_Quaternion.jl
```diff
@@ -67,22 +67,79 @@ let # test rotations
...
+    @test inv(q) * q ≈ q * inv(q) ≈ one(q)
+    @test sqrt(q) * sqrt(q) ≈ q
+    @test exp(log(q)) ≈ q
+    @test exp(zero(q)) ≈ one(q)
+    @test log(one(q)) ≈ zero(q)
+    @test exp2(log2(q)) ≈ q
+    @test exp10(log10(q)) ≈ q
+    @test expm1(log1p(q)) ≈ q
+    @test all(sincos(q) .≈ (sin(q), cos(q)))
+    VERSION ≥ v"1.6" && @test all(sincospi(q) .≈ (sinpi(q), cospi(q)))
```
hyrodium, line 122: Same as above.
```suggestion
    @test all(sincospi(q) .≈ (sinpi(q), cospi(q)))
```

hyrodium (changes requested):

src/Quaternion.jl
sethaxen: We could, but IMO this should be handled outside this PR. Also, so far maintaining support for Julia 1.0 hasn't been an issue, so I don't think we need to rush dropping pre-1.6 support.

src/Quaternion.jl
sethaxen: Yes, #62 but not #29. But I'm hoping to merge #52 instead to fix this.

sethaxen:
For reference, this is the heuristic function I used to identify potentially complex analytic functions in Base:

```julia
using FiniteDifferences

fdm = central_fdm(5, 1)

function get_complex_analytic_funs(mod::Module, z=rand(ComplexF64); ignore_defaults=true)
    fnames = names(mod)
    return filter(fnames) do fname
        f = getproperty(mod, fname)
        f isa Union{DataType,UnionAll} && return false
        hasmethod(f, Tuple{ComplexF64}) || return false
        ignore_defaults && hasmethod(f, Tuple{Number}) && return false
        try
            f(z) isa Complex || return false
        catch e
            println("$f errored on input $z")
            return false
        end
        return is_analytic_at(f, z)
    end
end

function is_analytic_at(f, z)
    J = only(jacobian(fdm, f, z))
    # Cauchy–Riemann equations
    return J[1,1] ≈ J[2,2] && J[1, 2] ≈ -J[2, 1]
end
```

```julia
julia> get_complex_analytic_funs(Base) |> sort
27-element Vector{Symbol}:
 :acos
 :acosh
 :asin
 :asinh
 :atan
 :atanh
 :cis
 :cispi
 :cos
 :cosh
 :cospi
 :exp
 :exp10
 :exp2
 :expm1
 :log
 :log10
 :log1p
 :log2
 :round
 :sin
 :sinh
 :sinpi
 :sqrt
 :tan
 :tanh
 :√
```

`round` should be handled outside of this PR. The rest are extended here.

Of these, `cis` and `cispi` are the two that might be confusing for users and could be removed. Because this package chooses a particular embedding of the complex numbers in the quaternions, `exp(im*q)` is computeable, but this is not the same thing as the extension of `cis(z)` to the quaternions. In fact, the extension has an intimate relationship with the axis and angle of a rotation in the same way that `cis(z)` is related to a rotation.

sethaxen:
@hydrodium this is ready for another review.

test/test_Quaternion.jl
```diff
@@ -156,22 +156,78 @@ Base.:(/)(a::MyReal, b::Real) = a.val / b
...
+        sqrt, inv, exp, exp2, exp10, expm1, log, log2, log10, log1p, cis,
+        sin, cos, tan, asin, acos, atan, sinh, cosh, tanh, asinh, acosh, atanh,
+        csc, sec, cot, acsc, asec, acot, csch, sech, coth, acsch, asech, acoth,
+        sinpi, cospi,
+    ]
+    # since every quaternion is conjugate to a complex number,
+    # one can establish correctness as follows:
+    @testset for fun in unary_funs
+        for i in 1:100
+            c = randn(ComplexF64)
```
hyrodium, line 186:
```suggestion
    unary_funs = [
        sqrt, inv, exp, exp2, exp10, expm1, log, log2, log10, log1p, cis,
        sin, cos, tan, asin, acos, atan, sinh, cosh, tanh, asinh, acosh, atanh,
        csc, sec, cot, acsc, asec, acot, csch, sech, coth, acsch, asech, acoth,
        sinpi, cospi,
    ]
    # since every quaternion is conjugate to a complex number,
    # one can establish correctness as follows:
    @testset "Check $(fun) compatibility with complex number" for fun in unary_funs
        q, q2 = randn(Quaternion{Float64}, 2)
        for _ in 1:100
            c = randn(ComplexF64)
```
I think it would be better to move `q` and `q2` into `@testset` to make the scope narrower and easier to read.

test/test_Quaternion.jl
```diff
@@ -156,22 +156,78 @@ Base.:(/)(a::MyReal, b::Real) = a.val / b
...
+    end
+    @testset for (f, finv) in [(asin, acsc), (acos, asec), (atan, acot), (asinh, acsch), (acosh, asech), (atanh, acoth)]
+        @test f(q) ≈ finv(inv(q))
+    end
+    @test q^1.3 ≈ exp(1.3 * log(q))
+    @test q^7.8 ≈ exp(7.8 * log(q))
+    @test q^1.3f0 ≈ exp(1.3f0 * log(q))
+    @test q^7.8f0 ≈ exp(7.8f0 * log(q))
+    @test cis(q) ≈ exp(normalize(q - real(q)) * q)
+    VERSION ≥ v"1.6" && @test cispi(q) ≈ cis(π * q)
```
hyrodium, line 222:
```suggestion
    @testset "Check the properties of each analytic function" begin
        for _ in 1:100
            q = randn(Quaternion{Float64})
            @test inv(q) * q ≈ q * inv(q) ≈ one(q)
            @test sqrt(q) * sqrt(q) ≈ q
            @test exp(log(q)) ≈ q
            @test exp(zero(q)) ≈ one(q)
            @test log(one(q)) ≈ zero(q)
            @test exp2(log2(q)) ≈ q
            @test exp10(log10(q)) ≈ q
            @test expm1(log1p(q)) ≈ q
            @test sinpi(q) ≈ sin(π * q)
            @test cospi(q) ≈ cos(π * q)
            @test all(sincos(q) .≈ (sin(q), cos(q)))
            @test all(sincos(zero(q)) .≈ (sin(zero(q)), cos(zero(q))))
            if VERSION ≥ v"1.6"
                @test all(sincospi(q) .≈ (sinpi(q), cospi(q)))
                @test all(sincospi(zero(q)) .≈ (sinpi(zero(q)), cospi(zero(q))))
            end
            @test tan(q) ≈ cos(q) \ sin(q) ≈ sin(q) / cos(q)
            @test tanh(q) ≈ cosh(q) \ sinh(q) ≈ sinh(q) / cosh(q)
            @testset for (f, finv) in [(sin, csc), (cos, sec), (tan, cot), (sinh, csch), (cosh, sech), (tanh, coth)]
                @test f(q) ≈ inv(finv(q))
            end
            @testset for (f, finv) in [(asin, acsc), (acos, asec), (atan, acot), (asinh, acsch), (acosh, asech), (atanh, acoth)]
                @test f(q) ≈ finv(inv(q))
            end
            @test q^1.3 ≈ exp(1.3 * log(q))
            @test q^7.8 ≈ exp(7.8 * log(q))
            @test q^1.3f0 ≈ exp(1.3f0 * log(q))
            @test q^7.8f0 ≈ exp(7.8f0 * log(q))
            @test cis(q) ≈ exp(normalize(q - real(q)) * q)
            VERSION ≥ v"1.6" && @test cispi(q) ≈ cis(π * q)
        end
    end
```
Adding `@testset` and `for _ in 1:100` seems better.

src/Quaternion.jl
```diff
@@ -115,43 +115,115 @@ end
...
-        scale *= sin(th) / th
+    z = complex(s, a)
+    w = f(z)
+    wr, wi = reim(w)
+    scale = wi / a
+    norm = _isexpfun(f) && iszero(s)
+    if a > 0
+        return Quaternion(wr, scale * q.v1, scale * q.v2, scale * q.v3, norm)
+    else  # quaternion may be real or complex
+        return Quaternion(wr, oftype(scale, wi), zero(scale), zero(scale), norm)
```
hyrodium, line 150: Could you add more comments here? I think this corresponds to Pu(q)/|Pu(q)|, but it would be easier to understand if the special consideration of the first component is clearly stated.

hyrodium:
The `exp(::Quaternion)` is now much faster than current master branch.

**Current master**
```julia
julia> @benchmark exp(Quaternion(0,1,2,3))
BenchmarkTools.Trial: 10000 samples with 991 evaluations.
 Range (min … max):  41.754 ns … 75.420 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     43.341 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   43.569 ns ±  2.194 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

    ▆▂▁▁█▅                                                    ▁
  ▆▇██████▅▆▆▅▇▆▇▇▆▇▇▇▇▇▆▆▆▆▆▅▆▅▅▃▄▄▄▅▄▃▄▃▃▄▃▃▃▄▄▃▃▂▄▃▃▃▃▃▄▄▄ █
  41.8 ns      Histogram: log(frequency) by time      55.6 ns <

```

**With this PR**
```julia
julia> @benchmark exp(Quaternion(0,1,2,3))
BenchmarkTools.Trial: 10000 samples with 999 evaluations.
 Range (min … max):  11.564 ns … 350.632 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     11.894 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   12.043 ns ±   5.150 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                    █   ▇▇   █    ▂                             
  ▂▂▂▂▂▂▁▁▂▂▂▃▄▅▅▃▅██▅▆▄██▅▄▃█▄▃▂▃█▃▂▂▃▃▁▂▂▂▁▂▁▁▁▁▁▁▁▁▁▂▁▁▂▂▂▂ ▃
  11.6 ns         Histogram: frequency by time         12.4 ns <

```

test/test_Quaternion.jl
```diff
@@ -156,22 +156,78 @@ Base.:(/)(a::MyReal, b::Real) = a.val / b
...
+    @testset for (f, finv) in [(sin, csc), (cos, sec), (tan, cot), (sinh, csch), (cosh, sech), (tanh, coth)]
+        @test f(q) ≈ inv(finv(q))
+    end
+    @testset for (f, finv) in [(asin, acsc), (acos, asec), (atan, acot), (asinh, acsch), (acosh, asech), (atanh, acoth)]
+        @test f(q) ≈ finv(inv(q))
+    end
+    @test q^1.3 ≈ exp(1.3 * log(q))
+    @test q^7.8 ≈ exp(7.8 * log(q))
+    @test q^1.3f0 ≈ exp(1.3f0 * log(q))
+    @test q^7.8f0 ≈ exp(7.8f0 * log(q))
```
hyrodium, line 220: Could you add some tests for `^(::Quaternion, ::Quaternion)`? I think it would be like this:
```julia
@test Quaternion(exp(1),0,0,0)^Quaternion(0,0,π/2,0) ≈ Quaternion(0,0,1,0)
```

src/Quaternion.jl
sethaxen: Is this clearer now?

sethaxen:
> The `exp(::Quaternion)` is now much faster than current master branch 🚀

It is faster, but not quite that much. You need to declare the quaternion before the benchmark or wrap it in a dollar sign, otherwise the compiler has too much information and cheats.

```julia
julia> q = Quaternion(0,1,2,3)
Quaternion{Int64}(0, 1, 2, 3, false)

julia> @btime exp($q)  # master
  53.776 ns (0 allocations: 0 bytes)
QuaternionF64(-0.8252990620752587, -0.15092132721996449, -0.30184265443992897, -0.45276398165989346, true)

julia> @btime exp($q)  # this PR
  39.218 ns (0 allocations: 0 bytes)
QuaternionF64(-0.8252990620752587, -0.15092132721996449, -0.30184265443992897, -0.45276398165989346, true)
```

src/Quaternion.jl
hyrodium: Thanks for the updates! I think it's clear enough.

hyrodium (approved):
I commented on a few minor points. It seems okay to merge, so I have approved.

test/test_Quaternion.jl
```diff
@@ -227,35 +227,35 @@ end
...
-            @test exp(Quaternion(0, 2, 0, 0)) == Quaternion(cos(2), sin(2), 0, 0, true)
-            @test exp(Quaternion(0, 0, 2, 0)) == Quaternion(cos(2), 0, sin(2), 0, true)
-            @test exp(Quaternion(0, 0, 0, 2)) == Quaternion(cos(2), 0, 0, sin(2), true)
-
+            @test exp(Quaternion(0, 0, 0, 0)) === Quaternion(1., 0., 0., 0., true)
+            @test exp(Quaternion(2, 0, 0, 0)) === Quaternion(exp(2), 0, 0, 0, false)
+            @test exp(Quaternion(0, 2, 0, 0)) === Quaternion(cos(2), sin(2), 0, 0, true)
+            @test exp(Quaternion(0, 0, 2, 0)) === Quaternion(cos(2), 0, sin(2), 0, true)
+            @test exp(Quaternion(0, 0, 0, 2)) === Quaternion(cos(2), 0, 0, sin(2), true)
+
```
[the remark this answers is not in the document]
sethaxen: Which spaces? The newlines? These are present on master.

I agree that it would have been better to have a formatter when the package was started (though none existed then), but I'd rather not add one now. Reformatting the entire package makes git blames to track the history for a bit of code useless.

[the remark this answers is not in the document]
hyrodium: 8 white spaces are added in line 235.

image

We don't have the spaces in the current master branch.
https://github.com/JuliaGeometry/Quaternions.jl/blob/766add41cb246faed9fa6d403b2aace86a3303bb/test/test_Quaternion.jl#L250-L252

There also are 8 white spaces in lines 241, 247, 253, and 258.

image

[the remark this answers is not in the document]
hyrodium: >I'd rather not add one now.

I agree with that. It should not be in this PR.

Can I add an action for formatting with JuliaFormatter.jl? If this sounds good, I'll make a PR.
cf. https://github.com/julia-actions/julia-format/blob/master/workflows/format_check.yml

Note that the current master branch has 2 and 4 white spaces for indent, and that makes confusing for new contributers.
Formatting code may break git diff, but I think it should be done as soon as possible.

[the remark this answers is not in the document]
sethaxen: I'd say open a PR and let's see how it looks. My concern is that, especially with the indentation, the formatter will touch almost every line of code, which would not be ideal. I'm all for clean, consistent formatting, but being able to explore the history easily is also important.

Issue #62, opened 2022-02-19
Should `Complex` be convertible to `Quaternion`?


hyrodium:
We already have constructor and `convert` method from `Complex` to `Quaternion`.
```julia
julia> Quaternion(Complex(1,2))
Quaternion{Int64}(1, 2, 0, 0, false)

julia> convert(Quaternion, Complex(1,2))
Quaternion{Int64}(1, 2, 0, 0, false)
```
However, we still missing the following methods `+` and `*`.
```julia
julia> Quaternion(1,2,3,4) + Complex(1,2)
ERROR: MethodError: no method matching imag(::Complex{Int64})
You may have intended to import Base.imag
Closest candidates are:
  imag(::Quaternion) at ~/.julia/dev/Quaternions/src/Quaternion.jl:44
  imag(::Octonion) at ~/.julia/dev/Quaternions/src/Octonion.jl:48
Stacktrace:
 [1] convert(#unused#::Type{Quaternion{Int64}}, z::Complex{Int64})
   @ Quaternions ~/.julia/dev/Quaternions/src/Quaternion.jl:21
 [2] _promote
   @ ./promotion.jl:327 [inlined]
 [3] promote
   @ ./promotion.jl:350 [inlined]
 [4] +(x::Quaternion{Int64}, y::Complex{Int64})
   @ Base ./promotion.jl:379
 [5] top-level scope
   @ REPL[23]:1

julia> Quaternion(1,2,3,4) * Complex(1,2)
ERROR: MethodError: no method matching imag(::Complex{Int64})
You may have intended to import Base.imag
Closest candidates are:
  imag(::Quaternion) at ~/.julia/dev/Quaternions/src/Quaternion.jl:44
  imag(::Octonion) at ~/.julia/dev/Quaternions/src/Octonion.jl:48
Stacktrace:
 [1] convert(#unused#::Type{Quaternion{Int64}}, z::Complex{Int64})
   @ Quaternions ~/.julia/dev/Quaternions/src/Quaternion.jl:21
 [2] _promote
   @ ./promotion.jl:327 [inlined]
 [3] promote
   @ ./promotion.jl:350 [inlined]
 [4] *(x::Quaternion{Int64}, y::Complex{Int64})
   @ Base ./promotion.jl:380
 [5] top-level scope
   @ REPL[24]:1
```

I think we have the following choices to solve this problem.

* Remove the constructor and `convert` method
[...]

PR #65, opened 2022-02-21
Implement sylvester and lyap for quaternions

### LINKED ISSUE #20: sylvester solver
It might be nice to define a `Base.sylvester(a::Quaternion, b::Quaternion, c::Quaternion)` method that solves `a*x + x*b + c == 0` for a quaternion `x`.   (e.g. it turns out that this is [needed for `sqrtm`](JuliaLang/julia/pull/20214)).

This can be solved e.g. via a [4x4 real matrix problem](http://conf.uni-ruse.bg/bg/docs/cp09/6.1/6.1-8.pdf).

sethaxen:
Fixes #20. Instead of mapping to matrices, it uses the following approach:

```
ax + xb + c = 0                                               # (1)
|a|^2 x b' + a' x |b|^2 + a' c b' = 0                         # (2) left- and right-multiply (1) by a' and b', respectively
x b' + a' x (|b|/|a|)^2 + a' c b'/|a|^2 = 0                   # (3) assume |a|>0 and divide by |a|^2
(2 re(b) x + a + a' (|b|/|a|)^2) x + c + a' c b'/|a|^2 = 0    # (4) add (1) and (3), and collect terms
-(2 re(b) x + a + a' (|b|/|a|)^2) x = c + a' c b'/|a|^2       # (5) rearrange
x = -(2 re(b) x + a + a' (|b|/|a|)^2) \ (c + a' c b'/|a|^2)   # (6) solve for x
```

For greater stability, if `|b| > |a|`, then we divide by `|b|^2` in (3), and continue.

### DIFF
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -56,6 +56,7 @@ abs2(q::Quaternion) = q.s * q.s + q.v1 * q.v1 + q.v2 * q.v2 + q.v3 * q.v3
 inv(q::Quaternion) = q.norm ? conj(q) : conj(q) / abs2(q)
 
 isfinite(q::Quaternion) = q.norm ? true : (isfinite(q.s) && isfinite(q.v1) && isfinite(q.v2) && isfinite(q.v3))
+iszero(q::Quaternion) = ~q.norm & iszero(real(q)) & iszero(q.v1) & iszero(q.v2) & iszero(q.v3)
 
 function normalize(q::Quaternion)
     if (q.norm)
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -307,3 +307,23 @@ function slerp(qa::Quaternion{T}, qb::Quaternion{T}, t::T) where {T}
         qa.v3 * ratio_a + qm.v3 * ratio_b,
     )
 end
+
+function sylvester(a::Quaternion, b::Quaternion, c::Quaternion)
+    abs2a = abs2(a)
+    abs2b = abs2(b)
+    if abs2a > abs2b
+        d1 = -(2real(b) + a + conj(a) * (abs2b / abs2a))
+        x = d1 \ (c + (conj(a) * c * conj(b)) / abs2a)
+    else
+        d2 = -(2real(a) + b + conj(b) * (abs2a / abs2b))
+        x = (c + (conj(a) * c * conj(b)) / abs2b) / d2
+    end
+    iszero(abs2a) && iszero(abs2b) && iszero(c) && return zero(x)
+    return x
+end
+
+function lyap(a::Quaternion, c::Quaternion)
+    x = (c + a \ c * a) / -4real(a)
+    iszero(a) && iszero(c) && return zero(x)
+    return x
+end
--- a/test/test_Quaternion.jl
+++ b/test/test_Quaternion.jl
@@ -349,3 +349,18 @@ end
         end
     end
 end
+
+@testset "sylvester/lyap" begin
+    a, b, c = randn(QuaternionF64, 3)
+    x = sylvester(a, b, c)
+    @test a * x + x * b ≈ -c
+    @test iszero(sylvester(a, b, zero(c)))
+    @test sylvester(a, zero(b), c) ≈ a \ -c
+    @test sylvester(zero(a), b, c) ≈ -c / b
+    @test iszero(sylvester(zero(a), zero(b), zero(c)))
+
+    @test lyap(a, c) ≈ sylvester(a, a', c)
+    @test lyap(b, c) ≈ sylvester(b, b', c)
+    @test iszero(lyap(a, zero(c)))
+    @test iszero(lyap(zero(a), zero(c)))
+end
--- a/test/test_Quaternion.jl
+++ b/test/test_Quaternion.jl
@@ -102,6 +102,15 @@ end
     @test convert(Octonion{Float64},Octonion(1,2,3,4,5,6,7,8)) === Octonion(1.0,2.0,3.0,4.0,5.0,6.0,7.0,8.0)
 end
 
+@testset "iszero" begin
+    @test iszero(Quaternion(0.0, 0.0, 0.0, 0.0))
+    @test !iszero(Quaternion(1.0, 0.0, 0.0, 0.0))
+    @test !iszero(Quaternion(0.0, 1.0, 0.0, 0.0))
+    @test !iszero(Quaternion(0.0, 0.0, 1.0, 0.0))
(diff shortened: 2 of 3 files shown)


sethaxen:
@stevengj I know the original issue is very old, but would you like to take a look at this?

sethaxen:
@hyrodium would you like to review this as well?

src/Quaternion.jl
```diff
@@ -381,3 +383,23 @@ function slerp(qa::Quaternion{T}, qb::Quaternion{T}, t::T) where {T}
         qa.v3 * ratio_a + qm.v3 * ratio_b,
     )
 end
+
```
hyrodium, line 386: I found the following `MethodError`.
```julia
julia> sylvester(Quaternion(1,2,3,4),0,0)
ERROR: MethodError: no method matching sylvester(::Quaternion{Int64}, ::Int64, ::Int64)
Closest candidates are:
  sylvester(::Union{Real, Complex}, ::Union{Real, Complex}, ::Union{Real, Complex}) at ~/julia/julia-1.7.1/share/julia/stdlib/v1.7/LinearAlgebra/src/dense.jl:1608
  sylvester(::Quaternion, ::Quaternion, ::Quaternion) at ~/.julia/dev/Quaternions/src/Quaternion.jl:387
Stacktrace:
 [1] top-level scope
   @ REPL[3]:1
```

We can resolve this error by these promotion:
```suggestion

sylvester(a::Quaternion, b::Union{Real,Quaternion}, c::Union{Real,Quaternion}) = sylvester(promote(a,b,c)...)
sylvester(a::Union{Real,Quaternion}, b::Quaternion, c::Union{Real,Quaternion}) = sylvester(promote(a,b,c)...)
sylvester(a::Union{Real,Quaternion}, b::Union{Real,Quaternion}, c::Quaternion) = sylvester(promote(a,b,c)...)
```
Note that these promotions seems able to replaced with
```julia
sylvester(a::Union{Real,Quaternion}, b::Union{Real,Quaternion}, c::Union{Real,Quaternion}) = sylvester(promote(a,b,c)...)
```
but this may overwrite [the original method `sylvester(a::Real, b::Real, c::Real)`](https://github.com/JuliaLang/julia/blob/df2abc6d77a72bde86c41373488a515f55251374/stdlib/LinearAlgebra/src/dense.jl#L1617).

hyrodium: I'm not sure we need to deal with `Complex` number with the promotion. (cf #62)

src/Quaternion.jl
```diff
@@ -381,3 +383,23 @@ function slerp(qa::Quaternion{T}, qb::Quaternion{T}, t::T) where {T}
...
+function sylvester(a::Quaternion, b::Quaternion, c::Quaternion)
```
hyrodium, line 387: It seems the following result is not correct with type promotion.
```julia
julia> sylvester(one(Quaternion),one(Quaternion),one(Quaternion{Rational{Int}}))
QuaternionF64(-0.5, 0.0, 0.0, 0.0, false)
```
This can be solved with the following.
```suggestion
sylvester(a::Quaternion, b::Quaternion, c::Quaternion) = sylvester(promote(a,b,c)...)
function sylvester(a::Quaternion{T}, b::Quaternion{T}, c::Quaternion{T}) where {T<:Real}
```

src/Quaternion.jl
```diff
@@ -381,3 +383,23 @@ function slerp(qa::Quaternion{T}, qb::Quaternion{T}, t::T) where {T}
...
+        x = d1 \ (c + inva * c * conj(b))
+    else
+        invb = conj(b) / abs2b
+        d2 = -2real(a) - b - invb * abs2a
+        x = (c + conj(a) * c * invb) / d2
+    end
+    return x
+end
+
+function lyap(a::Quaternion, c::Quaternion)
```
hyrodium, line 402: Same as `sylvester`, it would be nice to have promotions.

test/test_Quaternion.jl
```diff
@@ -426,3 +443,27 @@ end
...
+        @test isnan(sylvester(zero(a), zero(b), c))
+
+        @test lyap(a, c) ≈ sylvester(a, a', c)
+        @test lyap(b, c) ≈ sylvester(b, b', c)
+        @test iszero(lyap(a, zero(c)))
+        @test isnan(lyap(zero(a), c))
+    end
+    @test isnan(sylvester(zero(QuaternionF64), zero(QuaternionF64), zero(QuaternionF64)))
+    @test isnan(lyap(zero(QuaternionF64), zero(QuaternionF64)))
+end
```
hyrodium, line 469:
```suggestion
    @test_throws DivideError sylvester(zero(Quaternion{Rational{Int}}), zero(Quaternion{Rational{Int}}), zero(Quaternion{Rational{Int}}))
    @test_throws DivideError lyap(zero(Quaternion{Rational{Int}}), zero(Quaternion{Rational{Int}}))
end
```

hyrodium (commented):
I found some improvements around type promotions.

test/test_Quaternion.jl
```diff
@@ -102,6 +102,23 @@ end
     @test convert(Octonion{Float64},Octonion(1,2,3,4,5,6,7,8)) === Octonion(1.0,2.0,3.0,4.0,5.0,6.0,7.0,8.0)
 end
 
+@testset "iszero" begin
+    @test iszero(Quaternion(0.0, 0.0, 0.0, 0.0))
+    @test !iszero(Quaternion(1.0, 0.0, 0.0, 0.0))
+    @test !iszero(Quaternion(0.0, 1.0, 0.0, 0.0))
+    @test !iszero(Quaternion(0.0, 0.0, 1.0, 0.0))
+    @test !iszero(Quaternion(0.0, 0.0, 0.0, 1.0))
+    @test !iszero(Quaternion(0.0, 0.0, 0.0, 0.0, true))
```
hyrodium, line 111: We still ignoring the `norm` flag in `isone`. Should this fixed in this PR?
```julia
julia> isone(Quaternion(1,0,0,0,false))
true
```

sethaxen: No, `iszero` was only added here because we needed it (when we ued it). I think a future PR would add `isone` and some of the other `isXXX` methods.

src/Quaternion.jl
sethaxen: The specified promotions left some ambiguities, and by encoding all the methods directly, we were able to simplify quite a bit. e.g. anytime `a` and `b` commute with `x` (i.e. are real), we can compute with many fewer operations.

hyrodium (approved):

PR #66, opened 2022-02-21
update exp tests for `norm` flag

hyrodium:
With #52 merged, the `==` method no longer checks the `norm` field. This PR fixes the problem in the test.


sethaxen (approved):
LGTM, thanks!

PR #71, opened 2022-02-23
Simplify CI

sethaxen:
The current CI workflow is a bit overkill. Nothing we do is system/architecture-specific, so it's wasteful to run tests on Windows and OS X. But also, we should run it on a schedule so if for some reason something breaks on master, we don't suddenly find out when we open a PR.

Lastly, this uses more of the convenient Julia-specific GitHub actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,8 +2,10 @@ name: CI
 on:
   push:
     branches: [master]
-    tags: ["*"]
+    tags: [v*]
   pull_request:
+  schedule:
+    - cron: '0 0 * * *'
 jobs:
   test:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -19,53 +21,26 @@ jobs:
...
-      - uses: julia-actions/setup-julia@v1
-        with:
-          version: '1'
-      - run: julia --project=docs -e '
-          using Pkg;
-          Pkg.develop(PackageSpec(; path=pwd()));
-          Pkg.instantiate();'
-      - run: julia --project=docs docs/make.jl
+      - uses: julia-actions/julia-buildpkg@latest
+      - uses: julia-actions/julia-docdeploy@latest
```
sethaxen, line 43:
```suggestion
      - uses: julia-actions/julia-buildpkg@v1
      - uses: julia-actions/julia-docdeploy@v1
```

hyrodium (approved):

.github/workflows/ci.yml
```diff
@@ -2,8 +2,10 @@ name: CI
 on:
   push:
     branches: [master]
-    tags: ["*"]
+    tags: [v*]
   pull_request:
+  schedule:
+    - cron: '0 0 * * *'
```
hyrodium, line 8: Do we need this scheduled ci? I was overlooking these lines in the previous review.

sethaxen: Need? No. Sometimes, though, changes to dependencies or Julia itself are made between PRs, causing the CI to fail inexplicably when someone next goes to open a PR. This mitigates that. Quaternions is simple enough that this is not likely though, but it doesn't harm to run the CI once a day.

hyrodium: Thanks for the quick response! I just thought updating `gh-pages` every day is a little too frequent.

PR #72, opened 2022-02-23
Reorganize tests

sethaxen:
The current tests are a mess, and much of the code isn't tested. This PR completely reorganizes the tests, leaving a lot of empty test sets to mark where tests are needed. It doesn't add any tests; these should be handled in future PRs.

Because this PR changes every line of the test code, it also runs JuliaFormatter on the test files with BlueStyle.


test/Octonion.jl
```diff
@@ -0,0 +1,146 @@
...
+        end
+    end
+
+    @testset "basic" begin
+        real
+        imag
+        conj
+        float
+        Quaternions.abs_imag
+    end
```
hyrodium, line 94: Is this forgotten to add `@test`?

test/Octonion.jl
```diff
@@ -0,0 +1,146 @@
+
```
hyrodium, line 1: Do we need this empty line?

test/Octonion.jl
```diff
@@ -0,0 +1,146 @@
...
+            test_group(q1, q2, q3, *, one(q), inv)
+            test_multiplicative(q1, q2, *, norm)
+
+            # complex embedding
+            test_multiplicative(c1, c2, *, Quaternion)
+            test_multiplicative(c1, c2, +, Quaternion)
+        end
+    end
+
+    @testset "isreal" begin end
```
hyrodium, line 117: Is this still WIP?

hyrodium (commented):
Is this still WIP?

test/Octonion.jl
sethaxen: I left these here to note which functions should be tested here, but I've replaced this with a #TODO comment

test/Octonion.jl
sethaxen: No, as noted in the OP, I left empty test sets to mark where tests should be added. The goal this PR is a pure restructure, not to add new tests.

hyrodium: Ah, sorry that I overlooked the description.

test/Octonion.jl
```diff
@@ -0,0 +1,141 @@
+using Quaternions
+using LinearAlgebra
+using Random
+using Test
+
+@testset "Quaternion" begin
```
hyrodium, line 6:
```suggestion
@testset "Octonion" begin
```

test/Quaternion.jl
```diff
@@ -0,0 +1,442 @@
+using Quaternions
+using LinearAlgebra
+using Random
+using Test
```
hyrodium, line 4: Do we need these `using` here? I thought this can be moved to `runtests.jl`.

hyrodium: Ah, I was always commenting out the `include` in `runtests.jl` for that. :smile:

hyrodium (approved):
>85.32% <0.00%> (-0.55%)

I'm not sure the reason for the change of the coverage rate, but we can add more tests in another PR.

EDIT: https://app.codecov.io/gh/JuliaGeometry/Quaternions.jl/compare/72/changes#D1L15

sethaxen:
> > 85.32% <0.00%> (-0.55%)
> 
> I'm not sure the reason for the change of the coverage rate, but we can add more tests in another PR.
> 
> EDIT: https://app.codecov.io/gh/JuliaGeometry/Quaternions.jl/compare/72/changes#D1L15

Yes, this is because `sample` was using this constructor, but it was never explicitly tested. All the constructors need tests in a later PR.

PR #73, opened 2022-02-24
Fix `qrotation(dcm::Matrix{T})`

### LINKED ISSUE #67: Issues around conversion DCM to Quaternion
A conversion from DCM (direction cosine matrix) to `Quaternion` is defined here.

https://github.com/JuliaGeometry/Quaternions.jl/blob/v0.4.8/src/Quaternion.jl#L236-L256

The function `qrotation` has the following problems:

* It only accepts `Matrix`. This should be replaced with `AbstractMatrix`.
* The returned `Quaternion` doesn't have the `norm` flag `true`.
* The returned quaternion might have rotation angle larger than π.

The last problem is same as https://github.com/JuliaGeometry/Rotations.jl/pull/97.

Here's a script to check the returned value with `MRP`.

```julia
using Quaternions
using Rotations
using Plots
using GeometryBasics
plotly()

rs = rand(RotMatrix{3},10000)

qs = Quaternions.qrotation.(Matrix.(rs))

ps = QuatRotation{Float64}[]
for q in qs
    push!(ps, QuatRotation{Float64}(Quaternion(q.s,q.v1,q.v2,q.v3,true)))
end

points = Point.(Rotations.params.(MRP.(ps)))
scatter(points)
```

https://user-images.githubusercontent.com/7488140/155090993-2ba9100d-4d2c-4722-ab08-3e9e1758a964.mp4

hyrodium:
This PR fixes #67.
I have resolved the second and third issue in the following list.

* It only accepts `Matrix`. This should be replaced with `AbstractMatrix`.
* The returned `Quaternion` doesn't have the `norm` flag `true`.
* The returned quaternion might have rotation angle larger than π.

The first one will be fixed by #35.


hyrodium:
The scatter plot is now a ball.

https://user-images.githubusercontent.com/7488140/155521758-5deb916f-6b19-4e87-8a07-707d6ea3b60a.mp4

hyrodium:
Can I add tests for this? I thought this may make conflicts with #72.

sethaxen:
> @sethaxen
> Can I add tests for this? I thought this may make conflicts with #72.

I think if you keep them in a well-marked testset, then it shouldn't be a problem to merge it in after #72.

sethaxen:
> The returned quaternion might have rotation angle larger than π.

Can you elaborate on what the problem here is and how this PR fixes it?

hyrodium:
@sethaxen 
Could you review this PR again? Sorry if my explanation is poor.

sethaxen:
> @sethaxen Could you review this PR again? Sorry if my explanation is poor.

Sorry for the delay in replying. It's still not clear to me what the problem with the current implementation is? Is it incorrect? Numerically unstable? Unexpected? And how does the new implementation resolve these issues?

hyrodium:
The mapping from rotation matrix to quaternion is not unique because the set of unit quaternions (≃ S³ ≃ SU(2)) is a double cover of SO(3). I think the most natural range of the mapping should be {q=w+ix+jy+kz | w²+x²+y²+z²=1, w≥0}. Note that I don't much care about the boundary {q=w+ix+jy+kz | w²+x²+y²+z²=1, w=0}.
The following are why I choose the set {q=w+ix+jy+kz | w²+x²+y²+z²=1, w≥0}.

* q and -q represents the same rotation, so we need to divide the sphere S³ into two hemispheres.
* w is the most special axis in (w,x,y,z)
* {w≥0} is better than {w≤0} because the former set includes w=1, multiplicative identity.

>Is it incorrect?

It depends on how we define "correct". If we do not distinguish antipodal points `q` and `-q`, then the current `master` implementation is correct. But, I feel we need to care about the mapping more.

>Numerically unstable?

No.

>Unexpected?

Yes. Because of [this `if` statement](https://github.com/JuliaGeometry/Quaternions.jl/blob/master/src/Quaternion.jl#L307-L319), the range of the mapping SO(3) → S³ seems weird. This can be visualized via [stereographic projection](https://en.wikipedia.org/wiki/Stereographic_projection), and the result is [the scatter plot](https://github.com/JuliaGeometry/Quaternions.jl/issues/67#issue-1146590747).

We may define the `qrotation` without the `if` statement like the following instead of the current implementation.
```julia
function qrotation(dcm::AbstractMatrix{T}) where {T<:Real}
    # See https://arxiv.org/pdf/math/0701759.pdf
    a2 = 1 + dcm[1,1] + dcm[2,2] + dcm[3,3]
    b2 = 1 + dcm[1,1] - dcm[2,2] - dcm[3,3]
    c2 = 1 - dcm[1,1] + dcm[2,2] - dcm[3,3]
    d2 = 1 - dcm[1,1] - dcm[2,2] + dcm[3,3]

    # Same as a2 >= max(b2, c2, d2) branch
    a = sqrt(a2)/2
    return Quaternion(a, (dcm[3,2]-dcm[2,3])/4a, (dcm[1,3]-dcm[3,1])/4a, (dcm[2,1]-dcm[1,2])/4a)
end
``` 
But this is undesirable from the standpoint of calculation precision, especially if the `dcm` has errors.
Currently, we don't have tests for precision, But Rotations.jl has the same [function](https://github.com/JuliaGeometry/Rotations.jl/blob/master/src/unitquaternion.jl#L92-L124) and its test. We can refer [the test](https://github.com/JuliaGeometry/Rotations.jl/blob/v1.3.0/test/rotation_tests.jl#L274-L284). (I haven't added it yet in this PR.)

>And how does the new implementation resolve these issues?

As I said above, this PR enforces the output of `qrotation` to have a non-negative real part. i.e. the range of the function will be {q=w+ix+jy+kz | w²+x²+y²+z²=1, w≥0}.

test/Quaternion.jl
```diff
@@ -522,7 +522,13 @@ Base.:(/)(a::MyReal, b::Real) = a.val / b
             for _ in 1:100
                 q1 = nquatrand()
                 q2 = qrotation(rotationmatrix(q1), q1)
+                q3 = qrotation(rotationmatrix(q1))
                 @test q1 ≈ q2
+                @test q2 === q3 || q2 === -q3
```
sethaxen, line 527: Why is this check included?

test/Quaternion.jl
```diff
@@ -522,7 +522,13 @@ Base.:(/)(a::MyReal, b::Real) = a.val / b
...
+                @test real(q3) ≥ 0
```
sethaxen, line 528: Have you confirmed that this fails before the changes in this PR?

sethaxen (commented):
Thanks! That's very clear, and I agree that the `w>0` convention is the most reasonable one to follow. Just a couple minor questions.

test/Quaternion.jl
hyrodium: The range of `qrotation(dcm)` is {w≥0}, but the range of `qrotation(dcm, q)` is not. This test confirms that `qrotation(dcm, q)` differs only in sign from `qrotation(dcm)`.

test/Quaternion.jl
hyrodium: Yes, the test failed on the current master branch.

sethaxen (approved):

PR #74, opened 2022-02-25
Add missing tests

sethaxen:
The goal of this PR is to get (close to) complete coverage by adding missing tests. for already-implemented functionality.

PR #75, opened 2022-02-25
Add isunit and change norm field to property

### LINKED ISSUE #60: Do we really need normalized-flag in `Quaternion`?
# The problems
The following are what I am concerned about.

- Someone just needs quaternion operations, without the flag.
- Performance disadvantage
- The flag is not well-documented, and some operations don’t work properly with the flag.

## Someone just needs quaternion operations, without the flag.

See [comments in Makie.jl](https://github.com/JuliaPlots/Makie.jl/blob/master/src/utilities/quaternions.jl#L1-L5) for example.

## Performance disadvantage

The following `MyQuaternion` is a quaternion without the flag. The benchmark shows `-5.49% => improvement (5.00% tolerance)`.
```julia

julia> struct MyQuaternion{T<:Real} <: Number
           s::T
           v1::T
           v2::T
           v3::T
       end

julia> Base.:*(q::MyQuaternion, w::MyQuaternion) = MyQuaternion(q.s * w.s - q.v1 * w.v1 - q.v2 * w.v2 - q.v3 * w.v3,
                                                      q.s * w.v1 + q.v1 * w.s + q.v2 * w.v3 - q.v3 * w.v2,
                                                      q.s * w.v2 - q.v1 * w.v3 + q.v2 * w.s + q.v3 * w.v1,
                                                      q.s * w.v3 + q.v1 * w.v2 - q.v2 * w.v1 + q.v3 * w.s)

julia> mq = MyQuaternion(1.0, 2.0, 3.0, 4.0)
MyQuaternion{Float64}(1.0, 2.0, 3.0, 4.0)

julia> q = Quaternion(1.0, 2.0, 3.0, 4.0)
Quaternion{Float64}(1.0, 2.0, 3.0, 4.0, false)

julia> a = @benchmark $mq*$mq
 Range (min … max):  2.793 ns … 31.009 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     4.260 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   4.229 ns ±  0.486 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                  ▁           ▃▃ ▆ ▄       ▆ █ █ ▅       ▂ ▁ ▂
  ▅▁▇▁▆▁▁▁▁▁▁▁▅▁█▁█▁▇▁▁▁▁▁▃▁▇▁██▁█▁█▁▁▁▁▁█▁█▁█▁█▁█▁▁▁▁▁█▁█▁█ █
  2.79 ns      Histogram: log(frequency) by time     4.75 ns <

julia> b = @benchmark $q*$q
 Range (min … max):  3.213 ns … 18.089 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     4.330 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   4.474 ns ±  0.401 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                 ▅ █           ▂              
  ▂▁▂▁▂▁▂▁▁▁▁▁▁▁▂▁▂▁▂▁▂▁▁▁▁▁▂▁▄▄▁█▁█▁▅▁▁▁▁▁▃▁▇▁█▁▇▁▃▁▁▁▁▁▂▁▂ ▃
  3.21 ns        Histogram: frequency by time        5.17 ns <

julia> judge(mean(a),mean(b))
BenchmarkTools.TrialJudgement: 
  time:   -5.49% => improvement (5.00% tolerance)
  memory: +0.00% => invariant (1.00% tolerance)
```

## The flag is not well-documented, and some operations don’t work properly with the flag.
See #36, #39, #51 and #59.

# How to fix these problems
* If you need unit quaternions for rotations, I think `Rotations.QuatRotation` would be helpful.
    * `Rotations.jl` package have a dependency on `Quaternions.jl`, and conversion `Quaterion` <-> `QuatRotation` is supported.
* If you do not prefer the `Rotations` package, adding a new type `Quaternions.UnitQuaternon` seems better, I guess.
    * `QuatRotaions` is a rotation parametrization with `Quaternion`, but it is regarded as `AbstractMatrix`. i.e. the antipodal quaternions are equal. (https://github.com/JuliaGeometry/Rotations.jl/issues/171)

sethaxen:
As suggested in https://github.com/JuliaGeometry/Quaternions.jl/issues/60#issuecomment-1050393987, this PR adds an exported function `isunit` for checking if a number is normalized. It also changes the `norm` field to a (deprecated) property, which just forwards to `isunit`. All constructors and shorthands that use take a boolean argument for whether the number is unit are also deprecated.

This is not a breaking change. However, it should wait until #74 is merged so that we are sure it doesn't break existing functionality.

This will close #60

### DIFF
--- a/src/Octonion.jl
+++ b/src/Octonion.jl
@@ -64,6 +64,8 @@ float(q::Octonion{T}) where T = convert(Octonion{float(T)}, q)
 abs2(o::Octonion) = o.s * o.s + o.v1 * o.v1 + o.v2 * o.v2  + o.v3 * o.v3 + o.v4 * o.v4 + o.v5 * o.v5 + o.v6 * o.v6 + o.v7 * o.v7
 inv(o::Octonion) = o.norm ? conj(o) : conj(o) / abs2(o)
 
+isunit(o::Octonion) = isone(abs2(o))
+
 function normalize(o::Octonion)
   if (o.norm)
     return o
--- a/src/Octonion.jl
+++ b/src/Octonion.jl
@@ -7,7 +7,6 @@ struct Octonion{T<:Real} <: Number
   v5::T
   v6::T
   v7::T
-  norm::Bool
 end
 
 Octonion(s::Real, v1::Real, v2::Real, v3::Real, v4::Real, v5::Real, v6::Real, v7::Real, n::Bool = false) =
@@ -40,6 +39,14 @@ octo(p, v1, v2, v3, v4, v5, v6, v7) = Octonion(p, v1, v2, v3, v4, v5, v6, v7)
 octo(p, v1, v2, v3, v4, v5, v6, v7, n) = Octonion(p, v1, v2, v3, v4, v5, v6, v7, n)
 octo(x) = Octonion(x)
 octo(s, a) = Octonion(s, a)
+
+function Base.getproperty(o::Octonion, k::Symbol)
+  if k === :norm
+      Base.depwarn("`o.norm` is deprecated. Please use `isunit(o)` instead.", Symbol("o.norm"))
+      return isunit(o)
+  end
+  return getfield(o, k)
+end
 
 function show(io::IO, o::Octonion)
   pm(x) = x < 0 ? " - $(-x)" : " + $x"
--- a/src/Octonion.jl
+++ b/src/Octonion.jl
@@ -9,11 +9,9 @@ struct Octonion{T<:Real} <: Number
   v7::T
 end
 
-Octonion(s::Real, v1::Real, v2::Real, v3::Real, v4::Real, v5::Real, v6::Real, v7::Real, n::Bool = false) =
-  Octonion(promote(s, v1, v2, v3, v4, v5, v6, v7)..., n)
-Octonion(x::Real) = Octonion(x, zero(x), zero(x), zero(x), zero(x), zero(x), zero(x), zero(x), isunit(x))
-Octonion(z::Complex) = Octonion(z.re, z.im, zero(z.re), zero(z.re), zero(z.re), zero(z.re), zero(z.re), zero(z.re), isunit(z))
-Octonion(q::Quaternion) = Octonion(q.s, q.v1, q.v2, q.v3, zero(q.s), zero(q.s), zero(q.s), zero(q.s), isunit(q))
+Octonion(x::Real) = Octonion(x, zero(x), zero(x), zero(x), zero(x), zero(x), zero(x), zero(x))
+Octonion(z::Complex) = Octonion(z.re, z.im, zero(z.re), zero(z.re), zero(z.re), zero(z.re), zero(z.re), zero(z.re))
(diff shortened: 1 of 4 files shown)


src/Quaternions.jl
```diff
@@ -8,9 +8,22 @@ module Quaternions
   import Base: rand, randn
   import LinearAlgebra: lyap, norm, normalize, sylvester
   using Random
+  using DualNumbers
 
   Base.@irrational INV_SQRT_EIGHT 0.3535533905932737622004 sqrt(big(0.125))
 
+  """
+      isunit(x)
```
sethaxen, line 16: This naming is most intuitive (we say a "unit quaternion"), but `issign` would be more consistent with the definition that `x == sign(x)`.

mikmoore:
I made a comment on #60 about the default tolerance for `issign` (or `isunit` or whatever we call it). Depending on where we use `issign`, I'd have different thoughts. If we leave shortcuts in `inv`, `normalize`, and other functions based on it, the tolerance should be quite strict (maybe `10eps(T)`). If we don't use it internally but just have it for user convenience, we can make it a bit looser. In either case, but especially the latter, we should have an optional argument to specify the tolerance.

Of course, I'd certainly support keeping this PR limited to a simple removal of `.isnorm` and deferring changes to shortcuts to a separate PR.

src/Octonion.jl
```diff
@@ -34,10 +33,27 @@ promote_rule(::Type{Octonion{T}}, ::Type{Quaternion{S}}) where {T <: Real, S <:
...
+)
+Base.@deprecate octo(p, v1, v2, v3, v4, v5, v6, v7, n) octo(p, v1, v2, v3, v4, v5, v6, v7)
+
+function Base.getproperty(o::Octonion, k::Symbol)
+  if k === :norm
+      Base.depwarn("`o.norm` is deprecated. Please use `isunit(o)` instead.", Symbol("o.norm"))
+      return isunit(o)
+  end
+  return getfield(o, k)
+end
```
hyrodium, line 55: We already have 4 spaces indents in Octonion.jl. (e.g. https://github.com/JuliaGeometry/Quaternions.jl/blob/v0.5.6/src/Octonion.jl#L104-L114)
I think it would be better if the new code has 4 space indents.

hyrodium: We have removed Octonion from this package.

hyrodium:
I think we were planning to release breaking releases twice.

* Change `norm` field to property (`q1.norm&q2.norm` does not imply `(q1*q2).norm` with this change.)
* Remove `norm` property (`q.norm` doesn't work with this change.)

How about combining these changes in one breaking release?
I thought it will be okay with adding a warning to `getproperty(::Quaternion, :norm)`, and then remove it.

hyrodium:
https://github.com/JuliaGeometry/Quaternions.jl/pull/108 will be merged soon, so I'll close this issue.

PR #77, opened 2022-03-02
Add `imag_part` which returns `Tuple`

### LINKED ISSUE #61: `Quaternions.imag` is slow with `Vector`
It seems better to have the return value with type `Tuple` instead of `Vector`.
```julia

julia> q = Quaternion(1,2,3,4)
Quaternion{Int64}(1, 2, 3, 4, false)

julia> function new_imag(q::Quaternion)
           (q.v1,q.v2,q.v3)
       end
new_imag (generic function with 1 method)

julia> Quaternions.imag(q)
3-element Vector{Int64}:
 2
 3
 4

julia> new_imag(q)
(2, 3, 4)

julia> @benchmark Quaternions.imag($q)

BenchmarkTools.Trial: 10000 samples with 997 evaluations.
 Range (min … max):  20.596 ns …  1.775 μs  ┊ GC (min … max): 0.00% … 98.12%
 Time  (median):     22.907 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   26.725 ns ± 65.403 ns  ┊ GC (mean ± σ):  9.71% ±  3.92%

    ▂▆██▇▅▃▂▂▁                             ▁▁▁▁               ▂
  ▄▆██████████▇███▇▇▇▆▆▅▅▇▆▄▆▇▇▆▆▆▅▅▄▅▅▅▅▆██████▇▇▆▅▅▅▅▄▅▄▆▇▇ █
  20.6 ns      Histogram: log(frequency) by time      44.4 ns <

 Memory estimate: 80 bytes, allocs estimate: 1.

julia> @benchmark new_imag($q)
 Range (min … max):  0.977 ns … 10.895 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     2.375 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.397 ns ±  0.232 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                            █                 
  ▂▁▁▁▁▁▁▁▁▁▂▁▂▁▁▂▁▁▁▁▁▁▁▁▁▂▁▂▁▁▃▁▁▁▁▁▁▁▂▁▅▁█▁▁▇▁▁▁▁▁▁▁▂▁▃▁▃ ▂
  0.977 ns       Histogram: frequency by time        2.86 ns <

```
Note that this makes a breaking change.

hyrodium:
This PR fixes #61.

* I used `Base.depwarn` instead of `Base.@deprecate` because `Base.@deprecate` will exports `imag` function.
* I haven't added test for `imag`. This will be fixed in #74. (https://github.com/JuliaGeometry/Quaternions.jl/pull/74/files#diff-f8bd602e6a5913c49c69277a6a19d3325b311bc9f06eff13b45f1a3262d477c9R150-R160)


src/Quaternion.jl
```diff
@@ -44,7 +44,11 @@ end
 
 real(::Type{Quaternion{T}}) where {T} = T
 real(q::Quaternion) = q.s
-imag(q::Quaternion) = [q.v1, q.v2, q.v3]
+imag_part(q::Quaternion) = (q.v1, q.v2, q.v3)
+function imag(q::Quaternion)
+    Base.depwarn("`imag(q::Quaternion)` is deprecated, use `collect(imag_part(q))` instead.", :imag)
+    collect(imag_part(q))
+end
```
sethaxen, line 51: You should be able to use `@deprecate` without exporting `imag` like this:
```julia
@deprecate imag(q::Quaternion) collect(imag_part(q)) false
```

sethaxen, line 51: You can use `@deprecate` without exporting `imag` like this:
```suggestion
@deprecate imag(q::Quaternion) collect(imag_part(q)) false
```

hyrodium: Oh, I didn't know the usage. thanks!

src/Octonion.jl
```diff
@@ -140,7 +141,7 @@ end
 function log(o::Octonion)
   o, a = normalizea(o)
   s = o.s
-  M = abs(Octonion(imag(o)))
+  M = abs(0, Octonion(imag_part(o)...))
```
sethaxen, line 144:
```suggestion
  M = abs(Octonion(0, imag_part(o)...))
```

sethaxen (changes requested):
Can you make sure all lines changed are covered by tests? I know these will likely be replaced by #74, but this way we make sure this PR isn't breaking.

src/Octonion.jl
hyrodium: Oooh, thanks!

hyrodium:
>Can you make sure all lines changed are covered by tests? I know these will likely be replaced by https://github.com/JuliaGeometry/Quaternions.jl/pull/74, but this way we make sure this PR isn't breaking.

I thought this PR can wait for #74, but okay, I'll add some tests!

sethaxen:
> I thought this PR can wait for #74, but okay, I'll add some tests!

That's fine too! I plan to finish #74 this weekend (only dual quaternion tests left)

hyrodium:
Okay, I'll wait for that! #73 as well.

test/Octonion.jl
```diff
@@ -169,6 +169,7 @@ using Test
         @test real(q) === q.s
         @test_throws MethodError imag(q)
         @test Quaternions.imag(q) == [q.v1, q.v2, q.v3, q.v4, q.v5, q.v6, q.v7]
```
sethaxen, line 171:
```suggestion
        @test @test_deprecated(Quaternions.imag(q)) == [q.v1, q.v2, q.v3, q.v4, q.v5, q.v6, q.v7]
```

test/Quaternion.jl
```diff
@@ -179,6 +179,7 @@ Base.:(/)(a::MyReal, b::Real) = a.val / b
         @test real(q) === q.s
         @test_throws MethodError imag(q)
         @test Quaternions.imag(q) == [q.v1, q.v2, q.v3]
```
sethaxen, line 181:
```suggestion
        @test @test_deprecated(Quaternions.imag(q)) == [q.v1, q.v2, q.v3]
```

sethaxen (commented):
A minor suggestion to add deprecation tests. Also, please increment the patch version.

sethaxen (approved):

Issue #70, opened 2022-02-23
Adding logo to the document


hyrodium:
I'm planning to add a logo based on [Cayley Q8 quaternion multiplication graph](https://en.wikipedia.org/wiki/File:Cayley_Q8_quaternion_multiplication_graph.svg).

![image](https://user-images.githubusercontent.com/7488140/155314993-0e948c37-489d-4644-a5b1-98759e3e5b8f.png)

* The left one ignores the -1 vertex. (We can also think of the vertex of -1 as the infinity point)
* The middle one draws -1 vertex as a circle
* The right one is a "complete" graph

I prefer the first one because simplicity is important in logo design.

@sethaxen Do you have any thoughts on this?

hyrodium:
Here's a screenshot of the document with the left logo.

![image](https://user-images.githubusercontent.com/7488140/155321000-06e560d3-92f9-4f09-a02b-850728258083.png)

sethaxen:
Looks very nice! By any chance, are you programmatically generating it? (if so, would be good to have the code for that in the PR as well).

I agree, simpler is better for logos, so I do prefer the left one. Although I'm wondering even what it would look like if you completely removed the cycles including -1 (shown as straight lines in the background). For each of the options, I'd like to see what they look like on dark backgrounds as well and also as a favicon.
Some general thoughts:
- You're using purple to denote the bases (nodes) and red, green, and blue to denote the cycles, but in the Julia logo, red, green, and purple form the circles together, and blue is the outlier. Can you make the bases blue?
- I wonder what it would look like if it was rotated so that the 3 cycles in the middle were laid out similarly to the 3 circles in the Julia logo.

hyrodium:
>Looks very nice! By any chance, are you programmatically generating it? (if so, would be good to have the code for that in the PR as well).

I was using inkscape, but I'm planning to remake it with Luxor.jl:+1:

>Although I'm wondering even what it would look like if you completely removed the cycles including -1 (shown as straight lines in the background). For each of the options, I'd like to see what they look like on dark backgrounds as well and also as a favicon.

Okay, I'll try them later!

>You're using purple to denote the bases (nodes) and red, green, and blue to denote the cycles, but in the Julia logo, red, green, and purple form the circles together, and blue is the outlier. Can you make the bases blue?

I agree that the blue color is an outlier, but I'm also thinking that the purple is the most special color in the Julia logo.

[...]

PR #80, opened 2022-03-12
Add logo for the documentation

hyrodium:
This PR adds a logo based on #70.


sethaxen:
I'd prefer to not programmatically generate the logo when building the docs for several reasons:
1. If something changes in a dependency, the logo then can change in appearance.
2. It's wasteful; more time is taken to build dependencies and create the logo when building docs that should almost always not produce a different result.
3. The executable `convert` is not available on all systems (e.g. mine), so this makes it impossible for users to build the docs on such systems without commenting out logo generation. We should avoid barriers to users contributing to the documentation.

I would prefer we include the script to generate the logo as an asset but also include the output of the script. Then we are free to re-run the script periodically whenever we want and have it in version control, but we avoid these issues.

hyrodium:
Thanks for the suggestion! I have removed the logo generation with `make.jl`.

sethaxen (approved):
LGTM!

PR #82, opened 2022-04-03
Reorder operations in `*`, `abs2`

mikmoore:
The purpose of this PR is to re-order the operations in `*`. The primary motivator was to ensure that `q' * q == q * q' == abs2(q)` exactly, even for float operations. The previous arithmetic was dependent on lucky rounding so these equalities were only approximate. I made sure that `Octonion` matches `Quaternion` for these operations when the back half is zero, so that promotion doesn't risk changing arithmetic. I also added methods for scalar multiplication (the compiler wasn't able to compile away the zeros on its own) and some basic tests to ensure that `*` doesn't get broken by accident if somebody tries to improve it in the future.

There is also some performance improvement due to spending some time fiddling with the ordering of operations. `*(::QuaternionF64,::QuaternionF64)` goes from 4ns down to 3ns on my machine (and with #75 it looks like it might drop further to the 2.5ns ballpark). In the process I tried to get `abs2` to be as fast as possible, since #75 might make use of it. I changed `abs` to track these changes.

There are alternative orderings that could permit some `fma` instructions without breaking the target equalities and presumably be more accurate (and slightly faster on machines with native `fma`), but when I experimented with `fma` versions they weren't generating faster code. Also, it's not easy to add "fma only if native" operations right now (`muladd` can re-associate adjacent operations, not just the arguments). Maybe somebody revisits this in the future, but the gains won't be huge.

### DIFF
--- a/test/Octonion.jl
+++ b/test/Octonion.jl
@@ -295,6 +295,106 @@ using Test
         @test isnan(octo(1, 2, 3, 4, 5, 6, 7, NaN))
     end
 
+    @testset "*" begin
+        # verify basic correctness
+        q0 = Octonion(1,0,0,0,0,0,0,0)
+        q1 = Octonion(0,1,0,0,0,0,0,0)
+        q2 = Octonion(0,0,1,0,0,0,0,0)
+        q3 = Octonion(0,0,0,1,0,0,0,0)
+        q4 = Octonion(0,0,0,0,1,0,0,0)
+        q5 = Octonion(0,0,0,0,0,1,0,0)
+        q6 = Octonion(0,0,0,0,0,0,1,0)
+        q7 = Octonion(0,0,0,0,0,0,0,1)
+        @test q0 * q0 == q0
+        @test q0 * q1 == q1
+        @test q0 * q2 == q2
+        @test q0 * q3 == q3
+        @test q0 * q4 == q4
+        @test q0 * q5 == q5
+        @test q0 * q6 == q6
+        @test q0 * q7 == q7
+        @test q1 * q0 == q1
+        @test q1 * q1 == -q0
+        @test q1 * q2 == q3
+        @test q1 * q3 == -q2
+        @test q1 * q4 == -q7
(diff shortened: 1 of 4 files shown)


test/Quaternion.jl
```diff
@@ -267,6 +267,30 @@ Base.:(/)(a::MyReal, b::Real) = a.val / b
...
+        @test qj * q1 == qj
+        @test qj * qi == -qk
+        @test qj * qj == -q1
+        @test qj * qk == qi
+        @test qk * q1 == qk
+        @test qk * qi == qj
+        @test qk * qj == -qi
+        @test qk * qk == -q1
+    end
+
```
hyrodium, line 293:
```suggestion

    @testset "abs2" for _ in 1:100, T in (Float16, Float32, Float64)
        q = rand(Quaternion{T})
        @test abs2(q) == q'*q
    end

```
It would be nice to have these tests for

>The primary motivator was to ensure that q' * q == q * q' == abs2(q) exactly, even for float operations.

mikmoore: I wasn't sure if we wanted to commit to this as a guarantee but I suppose there's no reason we wouldn't want it.

test/Octonion.jl
```diff
@@ -295,6 +295,106 @@ using Test
...
+            @test qj * qi == -qk
+            @test qj * qj == -q1
+            @test qj * qk == qi
+            @test qk * q1 == qk
+            @test qk * qi == qj
+            @test qk * qj == -qi
+            @test qk * qk == -q1
+        end
+    end
+
```
hyrodium, line 397:
```suggestion

    @testset "abs2" for _ in 1:100, T in (Float16, Float32, Float64)
        o = rand(Octonion{T})
        @test abs2(o) == o'*o
    end

```
This `@testset` is already passed in the current `master` branch, but it would be better with this testset as well as Quaternion.

hyrodium (approved):
LGTM! (I'll merge after the abs2 testset for Octonion is updated.)

PR #85, opened 2022-04-29
Fix test for `slerp`

hyrodium:
We have `θ = clamp(rand() * 3.5, deg2rad(5e-1), π)` in the test for `slerp`, but this can be replaced with `θ = rand() * π`. The original issue around numerical stability was already solved in #78.


hyrodium:
@sethaxen Could you review PRs I've sent? If you are busy, I'll merge them after a week.

sethaxen (approved):

PR #88, opened 2022-04-30
Add support for `reim`

hyrodium:
The method `Base.imag(q::Quaternion)` should not be defined because the output type must be changed, but `Base.reim(c::Complex)` is a `Tuple`, so `Base.reim(q::Quaternion)` can be defined with return type `Tuple`.

### DIFF
--- a/src/Octonion.jl
+++ b/src/Octonion.jl
@@ -45,6 +45,7 @@ end
 
 real(o::Octonion) = o.s
 imag_part(o::Octonion) = (o.v1, o.v2, o.v3, o.v4, o.v5, o.v6, o.v7)
+Base.reim(o::Octonion) = (o.s, o.v1, o.v2, o.v3, o.v4, o.v5, o.v6, o.v7)
 @deprecate imag(o::Octonion) collect(imag_part(o)) false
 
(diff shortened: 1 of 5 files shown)


hyrodium:
Oh, I just noticed `reim(a::Real)` is a `NTuple{2, Real}`.

```julia
julia> reim(3)
(3, 0)
```

Should we close this PR, or rename `Base.reim(q::Quaternion)` with something like `Quaternions.reim_quat(q::Quaternion)`?

sethaxen:
> Oh, I just noticed `reim(a::Real)` is a `NTuple{2, Real}`.
> 
> ```julia
> julia> reim(3)
> (3, 0)
> ```
> 
> Should we close this PR, or rename `Base.reim(q::Quaternion)` with something like `Quaternions.reim_quat(q::Quaternion)`?

I agree, I would prefer a different function name than `reim`, but let me ask, what's the benefit of including any such function at all? Are there places where it substantially simplifies/reduces the code?

hyrodium:
>but let me ask, what's the benefit of including any such function at all? Are there places where it substantially simplifies/reduces the code?

Slightly improvement, but we can reduce these two lines to one line.
https://github.com/JuliaGeometry/Rotations.jl/blob/524f61d48396f3f604e90d422917efc247b9f58a/src/unitquaternion.jl#L126-L127

When I made this PR, I thought adding a method to `reim` would be better than `imag_part` because `imag_part` is not a function in `Base`.
I now agree with avoiding adding method to `reim`, and It seems okay with `real` and `imag_part` instead of `reim`.

Issue #90, opened 2022-08-20
Split `Octonion` into another package?


hyrodium:
I think the practical uses of `Quaternion` and `Octonion` are totally different, and these types don't have to be provided in one package.

sethaxen:
I agree, but also, I don't see major benefits from splitting `Octonion` to its own package. If it's here, we can have a uniform API and implementation, and we support legacy code. I doubt its inclusion changes the load time of this package much.

Probably `DualQuaternion` could/should be its own package, as this is where our DualNumbers dependency comes from. Without this dependency, the package loads in ~12ms even with `Octonion`, but with this dependency the load time is ~130ms. And as noted in #79, `DualQuaternion` could maybe be delivered without any additional code by releasing the `Real` type constraint on `Quaternion`.

sethaxen:
It seems no GitHub-hosted packages use Octonions in package code. LinearMaps.jl uses it in their tests. This is a good argument for moving Octonions to its own package. Do octonions have geometric applications? Should an Octonions.jl package live in JuliaGeometry as well or moved to an org like JuliaMath?

hyrodium:
Currently, I don't have applications for `Octonion`. Should we ask this topic on Julia Discourse or Slack?

sethaxen:
As others commented on Slack, either JuliaMath or JuliaGeometry would be a good host org for this package, though JuliaGeometry might make more sense due to its historical connections to the Quaternions package. 

Neither of us have rights in these orgs, but we don't need to to make the package and register it; it's easy to migrate the package to an org after registration. I do think it's important to preserve the contribution history, which we can do with `git filter-branch`. I've done this before, so I can make the repo, add you as an admin on it, and register it. Then we can transfer it when we get the ear of someone with the necessary org admin rights.

hyrodium:
I see. I don't have experience with `git filter-branch`, so it would be nice to be invited to your repository:+1:
After creating the repo, we can add warnings to `Octonion` constructor, and remove it in the next breaking release.

sethaxen:
I've created the repo at https://github.com/sethaxen/Octonions.jl and invited you to be a collaborator.

sethaxen:
I registered the package as-is to start the clock for the 3-day embargo: https://github.com/JuliaRegistries/General/pull/68578. We can make any other changes we like after the initial release.

Issue #86, opened 2022-04-29
Drop inefficient functions for rotation?


hyrodium:
The following functions return `Array` which is less efficient than `StaticArrays.StaticArray`.

* `rotationmatrix`
* `rotationmatrix_normalized`
* `axis`

These can be replaced with efficient functions in [`Rotations.jl`](https://github.com/JuliaGeometry/Rotations.jl) package. Adding warning messages to replace these functions with `Rotations.jl` package, and removing them in the next breaking release would be better, I guess.

Both `Quaternions.jl` and `Rotations.jl` packages are maintained under JuliaGeometry organization, so I think this change will not be problematic.

```julia
julia> using Quaternions, Rotations

julia> q = nquatrand()
QuaternionF64(0.09045921662676236, 0.9292679982955513, -0.0983890390919598, 0.34438018882858595, true)

julia> axis(q)
3-element Vector{Float64}:
  0.9330935318067537
 -0.09879407893716233
  0.34579790466015914

julia> rotation_axis(QuatRotation(q))
3-element StaticArrays.SVector{3, Float64} with indices SOneTo(3):
  0.9330935318067537
 -0.09879407893716233
  0.3457979046601592

julia> rotationmatrix(q)
3×3 Matrix{Float64}:
  0.743444  -0.245164   0.622243
 -0.120555  -0.964273  -0.235888
  0.657843   0.100355  -0.746439

julia> QuatRotation(q)
3×3 QuatRotation{Float64} with indices SOneTo(3)×SOneTo(3)(QuaternionF64(0.0904592, 0.929268, -0.098389, 0.34438, true)):
  0.743444  -0.245164   0.622243
 -0.120555  -0.964273  -0.235888
  0.657843   0.100355  -0.746439
```

hyrodium:
`angle` is not that inefficient, but I think `Base.angle(::Quaternion)` should be removed for consistency.

```julia
julia> using Quaternions

julia> angle(1+im)
0.7853981633974483

julia> angle(Quaternion(1+im))  # This issue is related to https://github.com/JuliaGeometry/Quaternions.jl/issues/62
1.5707963267948966

julia> angle(-1)
3.141592653589793

julia> angle(Quaternion(-1,0,0,0))  # This issue is not related to complex-quaternion compatibility
6.283185307179586

julia> -1 == Quaternion(-1,0,0,0)
true
```

hyrodium:
[...]

PR #92, opened 2022-09-16
Remove DualQuaternion

sethaxen:
This PR removes `DualQuaternion` and the dependency on DualNumbers. It instead adds an example to the documentation of using ForwardDiff with Quaternions to get the same functionality. The code in that example is actually the first steps one could take to define a package with just such utilities for dual quaternions using these 2 packages.

This is a breaking change, but we may want to make some more breaking changes before releasing, so the version number is marked `-DEV`.


hyrodium:
I think it would be better to release v0.5.7 with `Base.depwarn` (or `Base.@deprecate`). The next breaking release with v0.6.0 includes the following breaking changes, so it would be helpful to add detailed deprecated messages.

* Remove `linpol` (This function is already deprecated with `Base.@deprecate`)
* Remove `imag` (This function is already deprecated with `Base.@deprecate`)
* Remove `DualQuaternion` (I'm now making a PR to add deprecated messages)
* Remove `cis/cispi` #76 (This doesn't have deprecated messages.)

I'm not sure the following breaking changes will be included in v0.6.0.

* Remove `Octonion` (This should be done after creating Octonions.jl package, maybe in JuliaGeometry?)
* Remove rotation-related functions #86
* Remove `Complex`-`Quaternion` compatibility #62
* Change `norm` field to property #75 (`q.norm == true` does not lead `(q^2).norm == true` after this PR. It seems better to regard this PR as breaking changes, just in case.)

docs/src/examples/dual_quaternions.md
```diff
@@ -0,0 +1,159 @@
...
+isapprox(real(primal(x)' * tangent(x)), 0; atol=1e-10)
+```
+
+Here's how we use dual quaternions to transform a point:
+
+```@example dualquat
+p = randn(3)
+```
+
+```@example
```
hyrodium, line 129:
```suggestion
```@example dualquat
```

docs/src/examples/dual_quaternions.md
```diff
@@ -0,0 +1,159 @@
...
+    T[1:3, 1:3] .= R
+    T[1:3, 4] .= t
+    T[4, 1:3] .= 0
+    T[4, 4] = 1
+    return T
+end
+
+randdualquat(rng::AbstractRNG,T=Float64) = dualquat(rand(rng, Quaternion{T}), rand(rng, Quaternion{T}))
+randdualquat(T=Float64) = randdualquat(Random.GLOBAL_RNG,T)
+```
```
hyrodium, line 107:
````suggestion
randdualquat(T=Float64) = randdualquat(Random.GLOBAL_RNG,T)
nothing  # hide
```
````
It seems we don't need the information output of `randdualquat` here.

File: docs/src/examples/dual_quaternions.md [hunk 3]
```diff
@@ -0,0 +1,159 @@
+# Dual quaternions
+
+## Introduction
+
+The [dual quaternions](https://en.wikipedia.org/wiki/Dual_quaternion) are an example of "biquaternions."
```
hyrodium, line 5:
```suggestion
The [dual quaternions](https://en.wikipedia.org/wiki/Dual_quaternion) are an example of "biquaternions".
```

docs/src/examples/dual_quaternions.md
```diff
@@ -0,0 +1,159 @@
...
+end
+
+function tangent(d::DualQuaternion)
+    return Quaternion(
+        ForwardDiff.partials(real(d), 1),
+        ForwardDiff.partials.(imag_part(d), 1)...,
+    )
+end
+
+function Base.conj(d::DualQuaternion, type::Int)
```
hyrodium, line 66: `Base.conj` doesn't support this kind of 2nd argument, so I think it would be better to define a function with a new name.

hyrodium (commented):
Great! Thanks for the PR. I have added minor suggestions.

docs/src/examples/dual_quaternions.md
sethaxen: It's a silly rule, but periods are supposed to go within quotes like this.

sethaxen:
> I think it would be better to release v0.5.7 with `Base.depwarn` (or `Base.@deprecate`). The next breaking release with v0.6.0 includes the following breaking changes, so it would be helpful to add detailed deprecated messages.
> 
>     * Remove `linpol` (This function is already deprecated with `Base.@deprecate`)
> 
>     * Remove `imag` (This function is already deprecated with `Base.@deprecate`)
> 
>     * Remove `DualQuaternion` (I'm now making a PR to add deprecated messages)
> 
>     * Remove `cis/cispi` [Remove cis/cispi #76](https://github.com/JuliaGeometry/Quaternions.jl/pull/76) (This doesn't have deprecated messages.)

Good idea!


> I'm not sure the following breaking changes will be included in v0.6.0.
>
> 
>     * Remove `Octonion` (This should be done after creating Octonions.jl package, maybe in JuliaGeometry?)

I'll comment on this in #90.

>     * Remove rotation-related functions [Drop inefficient functions for rotation? #86](https://github.com/JuliaGeometry/Quaternions.jl/issues/86)

This seems fine to be in a later breaking release.

>     * Remove `Complex`-`Quaternion` compatibility [Should `Complex` be convertible to `Quaternion`? #62](https://github.com/JuliaGeometry/Quaternions.jl/issues/62)

This will be a little easier to do once this package contains just `Quaternion`, although it wouldn't be too much of a burden to do in v0.6.0. Why do you think it should wait?

>     * Change `norm` field to property [Add isunit and change norm field to property #75](https://github.com/JuliaGeometry/Quaternions.jl/pull/75) (`q.norm == true` does not lead `(q^2).norm == true` after this PR. It seems better to regard this PR as breaking changes, just in case.)

Yeah, dropping this is a priority, but it's a lot easier after `Quaternion` and `DualQuaternion` are gone. It _could_ be in v0.6.0, but that already seems busy, so it can wait.

docs/src/examples/dual_quaternions.md
hyrodium: Oh, sorry. The quoted thing is just one word but not a sentence, so I thought that was just a typo:sweat_smile:

hyrodium:
>This will be a little easier to do once this package contains just Quaternion, although it wouldn't be too much of a burden to do in v0.6.0. Why do you think it should wait?

I was just concerned about the release of `v0.6.0` will be delayed. It's before `v1.0.0`, so we can split the breaking changes.

>Yeah, dropping this is a priority, but it's a lot easier after `Quaternion` and `DualQuaternion` are gone. It could be in v0.6.0, but that already seems busy, so it can wait.

I agree with that. (`Quaternion` is a typo and should be `Octonion`? :smile:)

hyrodium:
The description should be updated, but I don't have permission for that.

image

sethaxen:
Good point, I've reached out to request I be upgraded to admin.

docs/src/examples/dual_quaternions.md
```diff
@@ -0,0 +1,154 @@
...
+end
+
+function tangent(d::DualQuaternion)
+    return Quaternion(
+        ForwardDiff.partials(real(d), 1),
+        ForwardDiff.partials.(imag_part(d), 1)...,
+    )
+end
+
+dualconj(d::DualQuaternion) = dualquat(conj(primal(d)), quat(-de.s, imag_part(de)...))
```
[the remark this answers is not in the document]
sethaxen: Thanks! I think rather I'd prefer something like this:
```suggestion
function dualconj(d::DualQuaternion)
    de = tangent(d)
    return dualquat(conj(primal(d)), quat(-real(de), imag_part(de)...))
end
```

sethaxen:
This PR should be ready for final review.

docs/src/examples/dual_quaternions.md
```diff
@@ -0,0 +1,157 @@
...
+!!! note
+    Previously this package contained a specialized `DualQuaternion` type.
+    This was removed in v0.6.0 because offered nothing extra over composing [ForwardDiff](https://github.com/JuliaDiff/ForwardDiff.jl) and Quaternions.
+
+## Utility functions
+
+First let's load the packages:
+
+```@example dualquat
+using ForwardDiff, Quaternions, Random
```
hyrodium, line 31:
```suggestion
using Quaternions, ForwardDiff, Random
```

I prefer that `Quaternions` comes first.

docs/src/examples/dual_quaternions.md
```diff
@@ -0,0 +1,157 @@
...
+```@example dualquat
+x = sign(randdualquat())
+```
+
+`sign(q) == q / abs(q)` both normalizes the primal part of the dual quaternion and makes the tangent part perpendicular to it.
+
+```@example dualquat
+abs(primal(x)) ≈ 1
+isapprox(real(primal(x)' * tangent(x)), 0; atol=1e-10)
+```
```
hyrodium, line 119:
````suggestion
```@repl dualquat
x = sign(randdualquat())
```

`sign(q) == q / abs(q)` both normalizes the primal part of the dual quaternion and makes the tangent part perpendicular to it.

```@repl dualquat
abs(primal(x)) ≈ 1
isapprox(real(primal(x)' * tangent(x)), 0; atol=1e-10)
```
````

I think `@repl` blocks are better than `@example` blocks here because we only have one output with `@example` block.

image

With the suggestion, the output will be like this:

image

File: docs/src/examples/dual_quaternions.md [hunk 8]
```diff
@@ -0,0 +1,157 @@
...
+
+Here's how we use dual quaternions to transform a point:
+
+```@example dualquat
+p = randn(3)
+```
+
+```@example dualquat
+transform(x, p)
+```
```
hyrodium, line 129:
````suggestion
```@repl dualquat
p = randn(3)
```

```@repl dualquat
transform(x, p)
```
````

File: docs/src/examples/dual_quaternions.md [hunk 9]
```diff
@@ -0,0 +1,157 @@
...
+
+We can check that this gives the same result as transforming with an affine transformation matrix:
+
+```@example dualquat
+transform(x, p) ≈ (transformationmatrix(x) * vcat(p, 1))[1:3]
+```
+
```
hyrodium, line 136:
````suggestion

## Example: homomorphism from dual quaternions to matrices

```@repl dualquat
y = sign(randdualquat())
```

```@repl dualquat
X = transformationmatrix(x)
Y = transformationmatrix(y)
XY = transformationmatrix(x*y)
X*Y ≈ XY
```
````

I thought it will be better to have a homomorphism example.
If we have this, `y = sign(randdualquat())` definition will be duplicated, and it should be removed in the motion planning section.

hyrodium (commented):
Most of the example seems fine! Just added some suggestions.

hyrodium (approved):
LGTM!

sethaxen:
Thanks for the suggestions! I made a few more changes. Should be finished now.

hyrodium (approved):
Thanks for the updates!

PR #93, opened 2022-09-16
Add deprecated messages for `DualQuaternion`

hyrodium:
I think it would be better to add deprecated message for `DualQuaternion` before merging #92.

### DIFF
--- a/src/DualQuaternion.jl
+++ b/src/DualQuaternion.jl
@@ -4,14 +4,14 @@ struct DualQuaternion{T<:Real} <: Number
   q0::Quaternion{T}
   qe::Quaternion{T}
   norm::Bool
-  function DualQuaternion(q0::Quaternion{T}, qe::Quaternion{T}, norm::Bool) where T <: Real
-    Base.depwarn("`DualQuaternion` will be removed in the next breaking release. Use `Quaternion{ForwardDiff.Dual}` instead.", :DualQuaternion)
-    return new{T}(q0, qe, norm)
-  end
   function DualQuaternion{T}(q0::Quaternion, qe::Quaternion, norm::Bool) where T <: Real
-    Base.depwarn("`DualQuaternion` will be removed in the next breaking release. Use `Quaternion{ForwardDiff.Dual}` instead.", :DualQuaternion)
+    Base.depwarn("`DualQuaternion` is deprecated and will be removed in the next breaking release. Use `Quaternion{ForwardDiff.Dual}` instead.", :DualQuaternion)
     return new{T}(q0, qe, norm)
   end
+end
+
+function DualQuaternion(q0::Quaternion{T}, qe::Quaternion{T}, norm::Bool) where T <: Real
+  return DualQuaternion{T}(q0, qe, norm)
 end
 
 DualQuaternion{T}(dq::DualQuaternion) where {T<:Real} = DualQuaternion{T}(dq.q0, dq.qe, dq.norm)
--- a/src/DualQuaternion.jl
+++ b/src/DualQuaternion.jl
@@ -4,6 +4,14 @@ struct DualQuaternion{T<:Real} <: Number
   q0::Quaternion{T}
   qe::Quaternion{T}
   norm::Bool
+  function DualQuaternion(q0::Quaternion{T}, qe::Quaternion{T}, norm::Bool) where T <: Real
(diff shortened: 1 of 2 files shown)


src/DualQuaternion.jl
```diff
@@ -4,6 +4,14 @@ struct DualQuaternion{T<:Real} <: Number
   q0::Quaternion{T}
   qe::Quaternion{T}
   norm::Bool
+  function DualQuaternion{T}(q0::Quaternion, qe::Quaternion, norm::Bool) where T <: Real
```
sethaxen, line 7: I think we need this to ensure the message is always raised:
```suggestion
  function DualQuaternion{T}(q0::Quaternion{T}, qe::Quaternion{T}, norm::Bool) where T <: Real
```

src/DualQuaternion.jl
```diff
@@ -4,6 +4,14 @@ struct DualQuaternion{T<:Real} <: Number
...
+    Base.depwarn("`DualQuaternion` is deprecated and will be removed in the next breaking release. Use `Quaternion{ForwardDiff.Dual}` instead.", :DualQuaternion)
+    return new{T}(q0, qe, norm)
+  end
+end
+
+function DualQuaternion(q0::Quaternion{T}, qe::Quaternion{T}, norm::Bool) where T <: Real
```
sethaxen, line 13: Why is this method necessary?

hyrodium: Without this method, `DualQuaternion(quat(1,2,3,4), quat(5,6,7,8), false)` throws `MethodError`. (On the current `master`, this doesn't throw errors.)

src/DualQuaternion.jl
hyrodium: On the current `master`, we have `DualQuaternion{T}(q0, qe, norm)` method that accepts `Any` type, and we should remove `::Quaternion` rather than adding type parameter `{T}`.

```julia
julia> methods(DualQuaternion{Float64})
# 10 methods for type constructor:
[1] (::Type{T})(x::AbstractChar) where T<:Union{AbstractChar, Number} in Base at char.jl:50
[2] (::Type{T})(x::Base.TwicePrecision) where T<:Number in Base at twiceprecision.jl:266
[3] DualQuaternion{T}(q0::Quaternion) where T<:Real in Quaternions at /home/hyrodium/.julia/dev/Quaternions/src/DualQuaternion.jl:17
[4] DualQuaternion{T}(dq::DualQuaternion) where T<:Real in Quaternions at /home/hyrodium/.julia/dev/Quaternions/src/DualQuaternion.jl:9
[5] DualQuaternion{T}(x::Real) where T<:Real in Quaternions at /home/hyrodium/.julia/dev/Quaternions/src/DualQuaternion.jl:26
[6] DualQuaternion{T}(d::DualNumbers.Dual) where T<:Real in Quaternions at /home/hyrodium/.julia/dev/Quaternions/src/DualQuaternion.jl:20
[7] DualQuaternion{T}(d1::DualNumbers.Dual, d2::DualNumbers.Dual, d3::DualNumbers.Dual, d4::DualNumbers.Dual) where T<:Real in Quaternions at /home/hyrodium/.julia/dev/Quaternions/src/DualQuaternion.jl:10
[8] DualQuaternion{T}(d1::DualNumbers.Dual, d2::DualNumbers.Dual, d3::DualNumbers.Dual, d4::DualNumbers.Dual, n::Bool) where T<:Real in Quaternions at /home/hyrodium/.julia/dev/Quaternions/src/DualQuaternion.jl:10
[9] (::Type{T})(x::T) where T<:Number in Core at boot.jl:772
[10] DualQuaternion{T}(q0, qe, norm) where T<:Real in Quaternions at /home/hyrodium/.julia/dev/Quaternions/src/DualQuaternion.jl:4
```

hyrodium: >I think we need this to ensure the message is always raised:

If we add `function` inside `struct`, the default constructor will be disabled. So there is no problem with the message?

hyrodium: I have removed the `::Quaterion` type limitation in the constructor function, and it now has the same methods as on the current `master`. I have confirmed these methods with

```julia
for s in sort(string.(methods(DualQuaternion))) println(s) end
for s in sort(string.(methods(DualQuaternionF64))) println(s) end
```

sethaxen (approved):
LGTM! Would you like to increment the patch version number and tag a new release?

hyrodium:
>Would you like to increment the patch version number and tag a new release?

I'll release `v0.5.7` after #99 is merged. That release will be the last release of `v0.5.x`

PR #95, opened 2022-09-16
update `quat(::Array)` to return `Array{<:Quaternion}`

hyrodium:
This PR fixes #94.
Note that this PR is considered as a breaking change.

https://github.com/JuliaGeometry/Quaternions.jl/blob/0be4cdd881872e14b480bdb69326f188f4b40236/test/Quaternion.jl#L112-L113

PR #97, opened 2022-09-17
Update documentation

hyrodium:
This PR adds `docs/src/api.md` page and adds more docstrings.

Screenshot:
image


docs/src/api.md
```diff
@@ -0,0 +1,25 @@
+# APIs
+
+```@docs
+Quaternion
+```
+
+```@docs
+quat
```
hyrodium, line 8: The docstring of `quat` will be added in #95.

docs/src/api.md
```diff
@@ -0,0 +1,25 @@
+# APIs
+
+```@docs
```
sethaxen, line 3: Is there a reason we can't use `@autodocs` here instead of listing the functions below?

src/Quaternion.jl
```diff
@@ -1,3 +1,13 @@
+"""
+    Quaternion{T<:Real} <: Number
+
+Quaternion number type with real and imaginary part of type `T`.
+
+`QuaternionF16`, `QuaternionF32` and `QuaternionF64` are aliases for
+`Quaternion{Float16}`, `Quaternion{Float32}` and `Quaternion{Float64}` respectively.
```
sethaxen, line 7:
```suggestion
`Quaternion{Float16}`, `Quaternion{Float32}`, and `Quaternion{Float64}`, respectively.
```

src/Quaternion.jl
```diff
@@ -29,15 +39,54 @@ quat(p, v1, v2, v3, n) = Quaternion(p, v1, v2, v3, n)
...
+
+Return the real part of the quaternion `q`.
+
+See also: [`imag_part`](@ref), [`quat`](@ref), [`isreal`](@ref).
+
+# Examples
+```jldoctest
+julia> real(quat(1,2,3,4))
+1
+```
```
sethaxen, line 53: Like in Base, I think we also want to capture the signature and examples of `real(::Type{<:Quaternion})` and `real(::Array{<:Quaternion})`

src/Quaternion.jl
```diff
@@ -29,15 +39,54 @@ quat(p, v1, v2, v3, n) = Quaternion(p, v1, v2, v3, n)
...
 real(::Type{Quaternion{T}}) where {T} = T
 real(q::Quaternion) = q.s
+
+"""
+    imag_part(q::Quaternion{T})
+
+Return the imaginary part of the quaternion `q` with type `Tuple{T,T,T}`.
+Note that this function is different from `Base.imag` which returns `Real`.
+
+See also: [`conj`](@ref).
```
sethaxen, line 64:
```suggestion
See also: [`real`](@ref), [`conj`](@ref).
```

sethaxen (changes requested):

src/Quaternion.jl
hyrodium: The docstring of `Complex` does not include that commas there.

https://github.com/JuliaLang/julia/blob/1ece3f64e093ba47bcd5e7a01193b8c9b05d441e/base/complex.jl#L8-L9

I have searched commas in JuliaLang/julia repository, and I found it seems better to put `,` before `respectively`, but I still don't understand why we need `,` before `and`.

docs/src/api.md
hyrodium: It's okay to use `@autodocs` here, but without the `@autodocs` we can choose the order of the docstrings and add more descriptions between docstrings in the future. If we missed some docstrings in md file, then Documenter.jl will raise warning about that. (`cis(::Quaternion)` will be removed, so I think we do not need the documentation.)

src/Quaternion.jl
sethaxen: The comma before `and` in a list (the Oxford comma) is entirely optional, but it should be consistently used or not used. I prefer it because it never reduces clarity, and there are some sentences that are not clear without it.

hyrodium: Ah, I see. thanks for the explanation! (and sorry for my English skill:sweat:)

src/Quaternion.jl
hyrodium: I added `real([quat(5,6,7,8), 9])` example to the docstring of `real(q::Quaternion)` because we don't define `real(A::AbstractArray)` and cannot add docstring to that.

README.md
```diff
@@ -7,57 +7,7 @@ A Julia implementation of quaternions.
 [![codecov](https://codecov.io/gh/JuliaGeometry/Quaternions.jl/branch/master/graph/badge.svg?token=dJBiR91dCD)](https://codecov.io/gh/JuliaGeometry/Quaternions.jl)
 
 [Quaternions](http://en.wikipedia.org/wiki/Quaternion) are best known for their suitability
-as representations of 3D rotational orientation. They can also be viewed as an extension of complex numbers.
+as representations of 3D rotational orientation.
+They can also be viewed as an extension of complex numbers.
 
-Implemented functions are:
```
hyrodium, line 12: I think this list is incomplete and should be moved to the documentation with Documenter.jl.

docs/src/api.md
sethaxen: okay, we can do this for now.

README.md
sethaxen: Agreed

src/Quaternion.jl
```diff
@@ -29,14 +39,72 @@ quat(p, v1, v2, v3, n) = Quaternion(p, v1, v2, v3, n)
 quat(x) = Quaternion(x)
 quat(s, a) = Quaternion(s, a)
 
+"""
+    real(T::Type)
```
sethaxen, line 43: This is too general. Let's do specifically
```suggestion
    real(T::Type{<:Quaternion})
```

src/Quaternion.jl
```diff
@@ -29,14 +39,72 @@ quat(p, v1, v2, v3, n) = Quaternion(p, v1, v2, v3, n)
...
+1
+
+julia> real([quat(5,6,7,8), 9])
+2-element Vector{Int64}:
+ 5
+ 9
+```
+"""
 real(q::Quaternion) = q.s
+
```
sethaxen, line 76: This is how we can document `real(::AbstractArray{<:Quaternion})` without defining a method:

```suggestion

"""
    real(A::AbstractArray{<:Quaternion})
    
Return an array containing the real part of each quaternion in `A`.

# Examples
````jldoctest
julia> real([quat(5,6,7,8), 9])
2-element Vector{Int64}:
 5
 9
````
"""
Base.real(::AbstractArray{<:Quaternion})

```

sethaxen (changes requested):

src/Quaternion.jl
hyrodium: I didn't know that. Thanks!

hyrodium:
Thanks for the suggestions! Should be ready for the final review.

docs/src/api.md
hyrodium: I added a temporary docstring in https://github.com/JuliaGeometry/Quaternions.jl/pull/97/commits/cffcb626b7ed3f8b74230f4ff071f9bd9cf29e21.

docs/src/api.md
```diff
@@ -0,0 +1,25 @@
+# APIs
```
sethaxen, line 1:
```suggestion
# API
```

src/Quaternion.jl
```diff
@@ -1,3 +1,13 @@
+"""
+    Quaternion{T<:Real} <: Number
+
+Quaternion number type with real and imaginary part of type `T`.
```
sethaxen, line 4:
```suggestion
Quaternion number type with real and imaginary parts of type `T`.
```

src/Quaternion.jl
```diff
@@ -1,3 +1,13 @@
...
+
+`QuaternionF16`, `QuaternionF32` and `QuaternionF64` are aliases for
```
sethaxen, line 6:
```suggestion
`QuaternionF16`, `QuaternionF32`, and `QuaternionF64` are aliases for
```

src/Quaternion.jl
```diff
@@ -24,19 +34,103 @@ Base.promote_rule(::Type{Quaternion{T}}, ::Type{S}) where {T <: Real, S <: Real}
...
+"""
+    quat(r, [i, j, k])
+
+Convert real numbers or arrays to quaternion. `i, j, k` defaults to zero.
+
+# Examples
+```jldoctest
+julia> quat(7)
+Quaternion{Int64}(7, 0, 0, 0, false)
+
```
sethaxen, line 46: Right now it reads like the above method only takes a real number and maybe an array. We could add an example of providing each part:
```suggestion

julia> quat(1.0, 2, 3, 4)
QuaternionF64(1.0, 2.0, 3.0, 4.0, false)

```

src/Quaternion.jl
```diff
@@ -24,19 +34,103 @@ Base.promote_rule(::Type{Quaternion{T}}, ::Type{S}) where {T <: Real, S <: Real}
...
+julia> quat([1, 2, 3])  # This output will be changed in the next breaking release for consistency. (#94)
```
sethaxen, line 47: Should we have a deprecation warning then?

src/Quaternion.jl
```diff
@@ -24,19 +34,103 @@ Base.promote_rule(::Type{Quaternion{T}}, ::Type{S}) where {T <: Real, S <: Real}
...
+julia> real([quat(5,6,7,8), 9])
+2-element Vector{Int64}:
+ 5
+ 9
+```
+"""
+Base.real(::AbstractArray{<:Quaternion})
+
+"""
+    imag_part(q::Quaternion{T})
```
sethaxen, line 104:
```suggestion
    imag_part(q::Quaternion{T}) -> NTuple{3, T}
```

src/Quaternion.jl
```diff
@@ -24,19 +34,103 @@ Base.promote_rule(::Type{Quaternion{T}}, ::Type{S}) where {T <: Real, S <: Real}
...
+
+Return the imaginary part of the quaternion `q` with type `Tuple{T,T,T}`.
```
sethaxen, line 106:
```suggestion
Return the imaginary part of the quaternion `q`.

```

src/Quaternion.jl
```diff
@@ -24,19 +34,103 @@ Base.promote_rule(::Type{Quaternion{T}}, ::Type{S}) where {T <: Real, S <: Real}
...
+Note that this function is different from `Base.imag` which returns `Real`.
```
sethaxen, line 107:
```suggestion
Note that this function is different from `Base.imag`, which returns `Real` for complex numbers.
```

src/Quaternion.jl
```diff
@@ -129,7 +223,7 @@ See Theorem 5 of [^Sudbery1970] for details.
 
 [^Sudbery1970]
```
sethaxen, line 224:
```suggestion
[^Sudbery1970]:
```

src/Quaternion.jl
```diff
@@ -24,19 +34,103 @@ Base.promote_rule(::Type{Quaternion{T}}, ::Type{S}) where {T <: Real, S <: Real}
...
+```
+"""
 Base.real(::Type{Quaternion{T}}) where {T} = T
+
+"""
+    real(q::Quaternion)
+
+Return the real part of the quaternion `q`.
+
+See also: [`imag_part`](@ref), [`quat`](@ref), [`isreal`](@ref).
```
sethaxen, line 78: We have no `isreal` docstring, and I don't think we need it.
```suggestion
See also: [`imag_part`](@ref), [`quat`](@ref)
```

sethaxen (changes requested):
Just a few minor tweaks. There are a few warnings printed when the docs build. e.g. two of the `real` docstrings are being excluded. You may need to explicitly list their signatures in the `@docs` block. Also, the dual quaternions example links to the `slerp` docs, but no such docstring exists.

src/Quaternion.jl
hyrodium: Agreed. I was just referring to the docstring of `Base.real`.

sethaxen (approved):
LGTM! Thanks!

PR #100, opened 2022-09-20
Remove `Octonion`

### LINKED ISSUE #90: Split `Octonion` into another package?

hyrodium:
This PR removes `Octonion` type. (closes #90)

I did not update the documentation not to make conflicts with #92.


sethaxen (commented):
Just a few more code deletions.

PR #106, opened 2022-10-06
Bump version to v0.6.0

hyrodium:
We have removed all of the deprecated structs and functions, so it's time to release `v0.6.0`.


sethaxen (approved):
Approving, but I think this should be the last PR merged before release. In particular, I'd like to see #97 merged first, since otherwise some links in the example docs are broken.

PR #108, opened 2022-10-19
Remove norm field and update normalization functions

### LINKED ISSUE #60: Do we really need normalized-flag in `Quaternion`?

### LINKED ISSUE #69: Drop normalize for quaternions
LinearAlgebra only implements 2 `normalize` methods:
```julia
julia> using LinearAlgebra

julia> methods(normalize)
# 2 methods for generic function "normalize":
[1] normalize(a::AbstractArray) in LinearAlgebra at /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/stdlib/v1.7/LinearAlgebra/src/generic.jl:1788
[2] normalize(a::AbstractArray, p::Real) in LinearAlgebra at /Applications/Julia-1.7.app/Contents/Resources/julia/share/julia/stdlib/v1.7/LinearAlgebra/src/generic.jl:1788
```

Notably, there is no `normalize(::Real)` or `normalize(::Complex)`. For good reason, as for numbers, this is the role of the `sign` function, which is defined generically for all numbers:
https://github.com/JuliaLang/julia/blob/bb2d8630f3aeb99c38c659a35ee9aa57bb71012a/base/number.jl#L161
We should deprecate `normalize` then and refer users to the `sign` function.

A difference between our `normalize` and `sign` functions, however, is how they treat zero quaternions:

```julia
julia> normalize(q)
QuaternionF64(0.1015814137657451, 0.8417214120379644, 0.5262271589168154, 0.0653548629598791, true)

julia> sign(q)
QuaternionF64(0.1015814137657451, 0.8417214120379644, 0.5262271589168154, 0.0653548629598791, false)

julia> normalize(zero(q))
QuaternionF64(NaN, NaN, NaN, NaN, true)

julia> sign(zero(q))
QuaternionF64(0.0, 0.0, 0.0, 0.0, false)
```

There are at least 4 different ways one could handle `normalize` for zero quaternions:
- return a 0 quaternion (`sign`)
- return NaNs (current behavior)
- return a random unit quaternion
- return a deterministic unit quaternion (e.g. `Quaternion(1)`)

I don't like the last 2, as they are arbitrary. The question is whether the current behavior is a bug or the desired behavior. If a bug, then we are free to define `Base.@deprecate normalize(x) sign(x)`. If desired, then we should define a different function for normalizing instead of overriding `LinearAlgebra.normalize` and deprecate `normalize`.

sethaxen:
This PR primarily removes the `:norm` field of `Quaternion` and updates all code and tests accordingly (fixes #60). It also:
- removes the `normalize` overload in favor of `sign`. (fixes #69)
- ~renames `normalizea` to `sign_abs` (exported)~
- removes `normalizea`
- removes `argq`
- removes `normalizeq` (only used internally for `argq` and makes no sense apart from it)

The first 2 are breaking. ~I wonder also if we should remove `sign_abs`. We don't use it internally, and something like `sign_abs` iss general-purpose and unrelated to quaternions.~

Supersedes #75

### DIFF
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -151,8 +151,9 @@ Base.isinf(q::Quaternion) = isinf(q.s) | isinf(q.v1) | isinf(q.v2) | isinf(q.v3)
 
 function sign_abs(q::Quaternion)
     a = abs(q)
-    q = q / a
-    (Quaternion(q.s, q.v1, q.v2, q.v3), a)
+    qnew = q / a
+    iszero(a) && return (zero(qnew), a)
+    return (qnew, a)
 end
 
 Base.@deprecate normalizea(q::Quaternion) sign_abs(q)
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -174,7 +174,15 @@ Base.isinf(q::Quaternion) = isinf(q.s) | isinf(q.v1) | isinf(q.v2) | isinf(q.v3)
 """
     sign(q::Quaternion) -> Quaternion
 
-Return zero if `q==0` and `q/|q|` otherwise.
+Return zero if `q==0` and ``q/|q|`` otherwise.
+
+# Examples
+```jldoctest
+julia> sign(Quaternion(4,0,0,0))
+QuaternionF64(1.0, 0.0, 0.0, 0.0)
+
+julia> sign(Quaternion(1,1,0,0))
+QuaternionF64(0.7071067811865475, 0.7071067811865475, 0.0, 0.0)
 """
 sign(::Quaternion)
 
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -47,15 +47,6 @@ function Base.promote_rule(::Type{Quaternion{T}}, ::Type{Complex{S}}) where {T <
 end
 Base.promote_rule(::Type{Quaternion{T}}, ::Type{Quaternion{S}}) where {T <: Real, S <: Real} = Quaternion{promote_type(T, S)}
 
-function Base.getproperty(q::Quaternion, s::Symbol)
-    if s === :norm
-        Base.depwarn("The `norm` field is deprecated and will be removed in the next breaking release (v0.7.0).", :Quaternion)
-        getfield(q,:norm)
-    else
-        getfield(q,s)
-    end
-end
-
 """
     quat(r, [i, j, k])
 
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -149,6 +149,14 @@ Base.iszero(q::Quaternion) = iszero(real(q)) & iszero(q.v1) & iszero(q.v2) & isz
 Base.isnan(q::Quaternion) = isnan(real(q)) | isnan(q.v1) | isnan(q.v2) | isnan(q.v3)
 Base.isinf(q::Quaternion) = isinf(q.s) | isinf(q.v1) | isinf(q.v2) | isinf(q.v3)
 
+# included strictly for documentation; the base implementation is sufficient
+"""
+    sign(q::Quaternion) -> Quaternion
+
+Return zero if `q==0` and `q/|q|` otherwise
+"""
+sign(::Quaternion)
+
 function sign_abs(q::Quaternion)
     a = abs(q)
     qnew = q / a
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -67,7 +67,6 @@ Quaternion{Int64}(0, 1, 2, 3)
 quat
 
 quat(p, v1, v2, v3) = Quaternion(p, v1, v2, v3)
-quat(p, v1, v2, v3, n) = Quaternion(p, v1, v2, v3, n)
 quat(x) = Quaternion(x)
 quat(s, a) = Quaternion(s, a)
 
@@ -103,7 +102,7 @@ Base.real(q::Quaternion) = q.s
 
 """
     real(A::AbstractArray{<:Quaternion})
-    
+
 Return an array containing the real part of each quaternion in `A`.
 
 # Examples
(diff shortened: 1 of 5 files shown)


hyrodium:
>I wonder also if we should remove sign_abs. We don't use it internally, and something like sign_abs iss general-purpose and unrelated to quaternions.

I agree with removing `sign_abs` (`normalizea`).

sethaxen:
For that matter, should we remove `argq`? It likewise is not part of the API and is nowhere used internally.

hyrodium:
>For that matter, should we remove `argq`? It likewise is not part of the API and is nowhere used internally.

I think the function can be removed too. I also think that the name of the function is not common and inappropriate.

src/Quaternion.jl
```diff
@@ -135,50 +133,31 @@ Compute the quaternion conjugate of a quaternion `q`.
...
-function normalizeq(q::Quaternion)
-    a = abs(q)
-    if a > 0
-        q = q / a
-        Quaternion(q.s, q.v1, q.v2, q.v3, true)
-    else
-        Quaternion(0.0, 1.0, 0.0, 0.0, true)
-    end
-end
+Return zero if `q==0` and `q/|q|` otherwise.
```
hyrodium, line 156: I think using Julia format
```suggestion
Return zero if `q==0` and `q/abs(q)` otherwise.
```
or math expression mode
```suggestion
Return zero if ``q=0`` and ``q/|q|`` otherwise.
```
would be better. If we follow the docstring of `Base.sign`, it will be like this:
```suggestion
Return zero if `q==0` and ``q/|q|`` otherwise.

# Examples
```jldoctest
julia> sign(Quaternion(4,0,0,0))
QuaternionF64(1.0, 0.0, 0.0, 0.0)

julia> sign(Quaternion(1,1,0,0))
QuaternionF64(0.7071067811865475, 0.7071067811865475, 0.0, 0.0)
```

hyrodium (commented):
Added a minor suggestion on a dosctring. I'll review other points after resolving the conflicts.

sethaxen:
@hyrodium this should be ready for final review.

hyrodium (approved):
Thank you for the updates! LTGM:+1:

PR #111, opened 2022-10-26
Remove rotation-related functions

### LINKED ISSUE #86: Drop inefficient functions for rotation?

hyrodium:
This PR removes rotation-related functions. (closes #86)
~Note that the documentation is failing to build because Rotations.jl assumes `Quaternion` has the `norm` field. I'd like to wait for https://github.com/JuliaGeometry/Rotations.jl/pull/243.~

Edit: The documentation does not depend on Rotations.jl now.

### DIFF
--- a/docs/src/examples/dual_quaternions.md
+++ b/docs/src/examples/dual_quaternions.md
@@ -88,8 +88,18 @@ function transform(d::DualQuaternion, p::AbstractVector)
     return pnew
 end
 
+function rotmatrix_from_quat(q::Quaternion)
+    sx, sy, sz = 2q.s * q.v1, 2q.s * q.v2, 2q.s * q.v3
+    xx, xy, xz = 2q.v1^2, 2q.v1 * q.v2, 2q.v1 * q.v3
+    yy, yz, zz = 2q.v2^2, 2q.v2 * q.v3, 2q.v3^2
+    r = [1 - (yy + zz)     xy - sz     xz + sy;
+            xy + sz   1 - (xx + zz)    yz - sx;
+            xz - sy      yz + sx  1 - (xx + yy)]
+    return r
+end
+
 function transformationmatrix(d::DualQuaternion)
-    R = rotationmatrix(rotation_part(d))
+    R = rotmatrix_from_quat(rotation_part(d))
     t = translation(d; first=false)
     T = similar(R, 4, 4)
     T[1:3, 1:3] .= R
--- a/docs/src/examples/rotations.md
+++ b/docs/src/examples/rotations.md
@@ -13,7 +13,7 @@ A 3D rotation can be represented by a [unit quaternion (versor)](https://en.wiki
 For example, a 90° rotation around the ``y``-axis is ``q = \frac{1}{\sqrt{2}} + 0i + \frac{1}{\sqrt{2}} j + 0k``.
 Rotations with quaternions have the following properties:
 
-* An additive inverse of a unit quaternion represents the same rotation.
+* The negative of a unit quaternion represents the same rotation.
 * A unit quaternion is more efficient for representing a rotation than a rotation matrix.
     * High computation accuracy
     * High computation performance, both in time and memory
--- a/docs/src/examples/rotations.md
+++ b/docs/src/examples/rotations.md
@@ -19,8 +19,8 @@ Rotations with quaternions have the following properties:
 * The conjugate of a unit quaternion represents the inverse rotation.
     * The quaternion has unit length, so conjugate and multiplicative inverse is the same.
 * The set of unit quaternion ``\left\{w + ix + jy + kz \in \mathbb{H} \ | \ x, y, z \in \mathbb{R} \right\} = U(1,\mathbb{H}) \simeq S^3`` forms a group, and the group is homomorphic to the following groups.
-    * ``SU(2) = \Set{R \in \mathcal{M}(2,\mathbb{C}) | R R^{*} = I}`` is isomorphic to ``U(1,\mathbb{H})``.
-    * ``SO(3) = \Set{R \in \mathcal{M}(3,\mathbb{R}) | R R^\top = I}`` is homomorphic to ``U(1,\mathbb{H})``, and the mapping ``U(1,\mathbb{H}) \to SO(3)`` is double covering.
+    * ``SU(2) = \{R \in \mathcal{M}(2,\mathbb{C}) \ | \ R R^{*} = I\}`` is isomorphic to ``U(1,\mathbb{H})``.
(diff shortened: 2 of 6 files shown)


docs/src/examples/rotations.md
```diff
@@ -0,0 +1,141 @@
...
+Rotations with quaternions have the following properties:
+
+* An additive inverse of a unit quaternion represents the same rotation.
+* Unit quaternion is more efficient to represent rotation than rotation matrix.
+    * High computation accuracy
+    * High computation performance, both in time and memory
+* A conjugate of a unit quaternion represents the inverse rotation.
+    * The quaternion has unit length, so conjugate and multiplicative inverse is the same.
+* The set of unit quaternion ``\left\{w + ix + jy + kz \in \mathbb{H} \ | \ x, y, z \in \mathbb{R} \right\} \simeq S^3`` is isomorphic to ``SU(2)``.
+    * These groups are homomorphic to ``SO(3)``.
```
sethaxen, line 23: We probably need to define SO(3). Can this and the below statement be merged?

docs/src/examples/rotations.md
```diff
@@ -0,0 +1,141 @@
...
+
+```@repl rotation
+q1 = quat_from_axisangle([0,1,0], deg2rad(90))  # 90° rotation around y-axis
+q2 = quat_from_axisangle([1,1,1], deg2rad(120))
+q3 = -q2  # additive inverse quaternion represents the same rotation
+```
+
+## Rotate a vector with a quaternion
+A vector ``v = (v_x, v_y, v_z)`` can be rotated by a unit quaternion ``q``.
+The rotated vector ``v' = (v_x', v_y', v_z')`` can be obtained as
```
sethaxen, line 52: What if we said `R(q)v` instead of `v'` to avoid confusion with `adjoint`?

docs/src/examples/dual_quaternions.md
```diff
@@ -88,8 +88,18 @@ function transform(d::DualQuaternion, p::AbstractVector)
     return pnew
 end
 
+function rotmatrix_from_quat(q::Quaternion)
```
sethaxen, line 91: Minor point, but I prefer we avoid `_from_quat` unless necessary, since in Julia, `_from_XXX` is handled by the type constraint in the method overload, so it's redundant.

sethaxen (changes requested):
I'll have more suggestions later, but here are a few.

docs/src/examples/rotations.md
```diff
@@ -0,0 +1,141 @@
...
+using Quaternions
+using LinearAlgebra
+```
+
+## Basics
+A 3D rotation can be represented by a [unit quaternion (versor)](https://en.wikipedia.org/wiki/Versor).
+For example, 90° rotation around y-axis is ``q = 1/\sqrt{2} + 0i + j/\sqrt{2} + 0k``.
+Rotations with quaternions have the following properties:
+
+* An additive inverse of a unit quaternion represents the same rotation.
```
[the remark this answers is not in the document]
hyrodium: Thank you for the suggestion! I have changed it that way.

docs/src/examples/rotations.md
```diff
@@ -0,0 +1,141 @@
...
+* Unit quaternion is more efficient to represent rotation than rotation matrix.
+    * High computation accuracy
+    * High computation performance, both in time and memory
+* A conjugate of a unit quaternion represents the inverse rotation.
+    * The quaternion has unit length, so conjugate and multiplicative inverse is the same.
+* The set of unit quaternion ``\left\{w + ix + jy + kz \in \mathbb{H} \ | \ x, y, z \in \mathbb{R} \right\} \simeq S^3`` is isomorphic to ``SU(2)``.
```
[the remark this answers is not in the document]
hyrodium: Thank you for the suggestion! I was really looking for the short symbol for the set of unit quaternions, and U(1, ℍ) looks great!

docs/src/examples/rotations.md
```diff
@@ -0,0 +1,141 @@
...
+See also [Rotations.jl documentation](https://juliageometry.github.io/Rotations.jl/stable/3d_quaternion/)
+
+```@example rotation
+using Quaternions
+using LinearAlgebra
+```
+
+## Basics
+A 3D rotation can be represented by a [unit quaternion (versor)](https://en.wikipedia.org/wiki/Versor).
+For example, 90° rotation around y-axis is ``q = 1/\sqrt{2} + 0i + j/\sqrt{2} + 0k``.
```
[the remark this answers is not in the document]
hyrodium: It seems the `\frac` expression is a little tight in the inline mode.

image

image

docs/src/examples/dual_quaternions.md
hyrodium: I agree with that as the common manner in the Julia language. However this is just documentation and readers cannot check the methods of the function by running `methods(XXX)`, and I thought explicit `XXX_from_YYY` is a little better.

docs/src/examples/rotations.md
hyrodium: Done!

hyrodium:
@sethaxen Thank you for the corrections, and sorry for my poor English:sweat_smile:. I think we can improve the documentation after the release of v0.7.0 if there are no major problems in this PR.

sethaxen (approved):

sethaxen:
> I think we can improve the documentation after the release of v0.7.0 if there are no major problems in this PR.

You're right, no problem at all! LGTM.

PR #116, opened 2022-12-07
Fix randomly failing log test on nightly

sethaxen:
The new tests for `log` added in #115 randomly fail on nightly due to an equality test. Here an approximate equality test is fine, so this PR uses that instead.

### DIFF
--- a/test/Quaternion.jl
+++ b/test/Quaternion.jl
@@ -379,10 +379,10 @@ end
             @testset "log" begin
                 @test log(zero(QuaternionF64)) === Quaternion(-Inf, 0, 0, 0)
                 @test log(one(QuaternionF64)) === Quaternion(0.0, 0, 0, 0)
-                @test log(-one(QuaternionF64)) == Quaternion(0.0, π, 0, 0)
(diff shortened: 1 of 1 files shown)


test/Quaternion.jl
```diff
@@ -379,10 +379,10 @@ end
             @testset "log" begin
                 @test log(zero(QuaternionF64)) === Quaternion(-Inf, 0, 0, 0)
                 @test log(one(QuaternionF64)) === Quaternion(0.0, 0, 0, 0)
-                @test log(-one(QuaternionF64)) == Quaternion(0.0, π, 0, 0)
+                @test log(-one(QuaternionF64)) ≈ Quaternion(0.0, π, 0, 0)
```
hyrodium, line 382: Very minor point, but the first argument doesn't have to be `Float64`.
```suggestion
                @test log(-one(QuaternionF64)) ≈ Quaternion(0, π, 0, 0)
```

hyrodium (approved):
LGTM! I have added a minor suggestion, but this PR can be merged without another approval.

PR #118, opened 2022-12-07
Remove angleaxis

sethaxen:
This method does not work and is not tested. It must have been missed in the PRs removing rotation-related functions


hyrodium (approved):
Thank you!

Issue #58, opened 2022-02-19
Add ChainRules rules for certain functions?


sethaxen:
It's becoming common in the ecosystem to add ChainRules rules for functions that use features that are problematic or slow for certain AD packages, like mutation, loops, or control flow. Maybe we should do this to, at least to add projections and such. https://github.com/JuliaDiff/ChainRules.jl/pull/540 should probably be finished first.

We already have ChainRulesCore as an indirect dependency via DualNumbers and then SpecialFunctions, so this would not make this package heavier.

sethaxen:
Since https://github.com/JuliaDiff/ChainRules.jl/pull/540 has stalled, we should probably just defensively use `ChainRulesCore.@opt_out` for functions that we implement and that we know depend on the quaternion product so that ChainRules-compatible ADs don't just silently do the wrong thing here.

PR #119, opened 2022-12-09
Add support for `RealDot`

hyrodium:
This PR adds `RealDot.realdot(::Quaternion, ::Quaternion)` as suggested in https://github.com/JuliaGeometry/Quaternions.jl/issues/58#issuecomment-1341678095.

### DIFF
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -321,7 +321,7 @@ true
     iszero(qb0) && throw(DomainError(qb0, "The input quaternion must be non-zero."))
     qa = qa0 / abs(qa0)
     qb = qb0 / abs(qb0)
-    coshalftheta = realdot(qa, qb)
+    coshalftheta = RealDot.realdot(qa, qb)
 
     if coshalftheta < 0
         qb = -qb
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -139,7 +139,7 @@ Base.conj(q::Quaternion) = Quaternion(q.s, -q.v1, -q.v2, -q.v3)
 Base.abs(q::Quaternion) = sqrt(abs2(q))
 Base.float(q::Quaternion{T}) where T = convert(Quaternion{float(T)}, q)
 abs_imag(q::Quaternion) = sqrt(q.v2 * q.v2 + (q.v1 * q.v1 + q.v3 * q.v3)) # ordered to match abs2
-Base.abs2(q::Quaternion) = (q.s * q.s + q.v2 * q.v2) + (q.v1 * q.v1 + q.v3 * q.v3)
+Base.abs2(q::Quaternion) = RealDot.realdot(q,q)
 Base.inv(q::Quaternion) = conj(q) / abs2(q)
 
 Base.isreal(q::Quaternion) = iszero(q.v1) & iszero(q.v2) & iszero(q.v3)
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -321,7 +321,7 @@ true
     iszero(qb0) && throw(DomainError(qb0, "The input quaternion must be non-zero."))
     qa = qa0 / abs(qa0)
     qb = qb0 / abs(qb0)
-    coshalftheta = qa.s * qb.s + qa.v1 * qb.v1 + qa.v2 * qb.v2 + qa.v3 * qb.v3
+    coshalftheta = realdot(qa, qb)
 
     if coshalftheta < 0
         qb = -qb
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -386,3 +386,6 @@ LinearAlgebra.lyap(a::Quaternion, c::Quaternion) = lyap(promote(a, c)...)
 # if a commutes with c, use a simpler expression
 LinearAlgebra.lyap(a::Real, c::Quaternion) = c / -2a
 LinearAlgebra.lyap(a::Quaternion, c::Real) = c / -2real(a)
+
+## RealDot
+@inline RealDot.realdot(p::Quaternion, q::Quaternion) = muladd(p.s, q.s, dot(imag_part(p), imag_part(q)))
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -388,4 +388,4 @@ LinearAlgebra.lyap(a::Real, c::Quaternion) = c / -2a
 LinearAlgebra.lyap(a::Quaternion, c::Real) = c / -2real(a)
 
 ## RealDot
-@inline RealDot.realdot(p::Quaternion, q::Quaternion) = muladd(p.s, q.s, dot(imag_part(p), imag_part(q)))
+@inline RealDot.realdot(p::Quaternion, q::Quaternion) = (p.s * q.s + p.v2 * q.v2) + (p.v1 * q.v1 + p.v3 * q.v3)
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -388,4 +388,5 @@ LinearAlgebra.lyap(a::Real, c::Quaternion) = c / -2a
 LinearAlgebra.lyap(a::Quaternion, c::Real) = c / -2real(a)
 
 ## RealDot
+# ordering chosen so that real(p'q) == real(q'p) == realdot(p, q) == realdot(q, p), i.e. exact equality
 @inline RealDot.realdot(p::Quaternion, q::Quaternion) = (p.s * q.s + p.v2 * q.v2) + (p.v1 * q.v1 + p.v3 * q.v3)
--- a/test/Quaternion.jl
+++ b/test/Quaternion.jl
@@ -576,8 +576,13 @@ end
         for _ in 1:10
             q1 = randn(Quaternion{T})
             q2 = randn(Quaternion{T})
+            # Check real∘dot is equal to realdot.
             @test real(dot(q1,q2)) == @inferred(realdot(q1,q2))
+            # Check realdot is commutative.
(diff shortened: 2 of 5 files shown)


src/Quaternion.jl
```diff
@@ -386,3 +386,6 @@ LinearAlgebra.lyap(a::Quaternion, c::Quaternion) = lyap(promote(a, c)...)
 # if a commutes with c, use a simpler expression
 LinearAlgebra.lyap(a::Real, c::Quaternion) = c / -2a
 LinearAlgebra.lyap(a::Quaternion, c::Real) = c / -2real(a)
+
+## RealDot
+@inline RealDot.realdot(p::Quaternion, q::Quaternion) = muladd(p.s, q.s, dot(imag_part(p), imag_part(q)))
```
sethaxen, line 391: This is almost as verbose as explicitly writing out the sum, so perhaps this is better?
```suggestion
@inline RealDot.realdot(p::Quaternion, q::Quaternion) = p.s * q.s + p.v1 * q.v1 + p.v2 * q.v2 + p.v3 * q.v3
```

test/Quaternion.jl
```diff
@@ -571,4 +571,13 @@ end
             @test_throws DivideError lyap(null, null)
         end
     end
+
+    @testset "RealDot with $T" for T in (Float32, Float64)
+        for _ in 1:10
+            q1 = randn(Quaternion{T})
+            q2 = randn(Quaternion{T})
+            @test real(dot(q1,q2)) ≈ realdot(q1,q2)
```
sethaxen, line 579:
```suggestion
            @test real(dot(q1,q2)) ≈ @inferred(realdot(q1,q2))
```

sethaxen (commented):
LGTM, just some minor suggestions.

There are 2 places in the code where we might consider using `realdot`:
```julia
Base.abs2(q::Quaternion) = RealDot.realdot(q, q)
# in slerp
coshalftheta = RealDot.realdot(qa, qb)
```

Merge with a patch version bump when you feel it's ready.

src/Quaternion.jl
hyrodium: I was understanding that `muladd` version is more efficient than the suggested explicit version, but the benchmark results are almost the same in my local machine.
Another reason for `muladd` was referring to `realdot(x::Complex, y::Complex)` method in RealDot.jl (https://github.com/JuliaMath/RealDot.jl/blob/6a8e18b23ac2d79d8337b272c78bca7f0ce51bf2/src/RealDot.jl#L17).

hyrodium:
>There are 2 places in the code where we might consider using `realdot`:

Replacing `abs2` causes failures with tests added by https://github.com/JuliaGeometry/Quaternions.jl/pull/82#discussion_r841934935 . We have the following choices to fix this problem.

* Keep the original `abs2` definition
* Replace `@test abs2(q) == q'*q` with `@test abs2(q) ≈ q'*q`
* Fix `realdot` definition (x-ref: https://github.com/JuliaGeometry/Quaternions.jl/pull/119#discussion_r1045101872)

Do you have any thoughts on this?

sethaxen:
> Fix `realdot` definition (x-ref:

I think this makes the most sense.

src/Quaternion.jl
sethaxen: I used to think `muladd` was faster too, and then IIRC someone corrected me. But now I don't remember. I've never seen a difference in benchmarks. I tend to use `muladd` unless it makes the code much less clear.

@mikmoore do you know anything about this?

mikmoore: `muladd` gives the compiler the *option* to use either `a*b+c` or `fma(a,b,c)`. `fma` is more accurate and, in isolation, is faster than the two distinct operations (unless it's not natively supported by the CPU). However, it is sometimes skipped because register shuffling or other things would be required. Note that inlining can change the circumstances and result in different choices as well.

mikmoore:
I am going to suggest that `realdot` be made to match the calculation of `real(p*q)` -- i.e., copy the first line of the `*` definition. I already spent significant effort in #82 making sure that `abs2(q)` and `real(q*q')` are *exactly* equal and as fast as I could manage. Anything but exact equality can cause headaches for users. Any speed difference will be negligible outside of microbenchmarks.

Also, make sure that `code_native((p,q)->real(p*q),NTuple{2,QuaternionF64})` doesn't succeed in dropping the imaginary parts of the calculation. If it does, then I would define both `realdot` and `abs2` in terms of `*` so that we don't need to worry about keeping the implementations synchronized any more. The compiler failed to do this a year ago, but it's worth trying again while we're here (having removed `.norm` might have helped).

test/Quaternion.jl
```diff
@@ -571,4 +571,13 @@ end
             @test_throws DivideError lyap(null, null)
         end
     end
+
+    @testset "RealDot with $T" for T in (Float32, Float64)
+        for _ in 1:10
+            q1 = randn(Quaternion{T})
+            q2 = randn(Quaternion{T})
+            @test real(dot(q1,q2)) ≈ @inferred(realdot(q1,q2))
```
mikmoore, line 579: If we implement `realdot` to follow `*`, this test can pass with `==`.

sethaxen:
> I am going to suggest that `realdot` be made to match the calculation of `real(p*q)` -- i.e., copy the first line of the `*` definition.

It should be `real(conj(p) * q)`, but otherwise, yes, I agree.

> Also, make sure that `code_native((p,q)->real(p*q),NTuple{2,QuaternionF64})` doesn't succeed in dropping the imaginary parts of the calculation. If it does, then I would define both `realdot` and `abs2` in terms of `*` so that we don't need to worry about keeping the implementations synchronized any more. The compiler failed to do this a year ago, but it's worth trying again while we're here (having removed the old `.norm` field might have helped).

Even if the compiler manages to not calculate the imaginary part, it's still worth having these overloads, because common code transformations, such as automatic differentiation still happen at an earlier step in compilation process; the resulting code is more complex, and the compiler might not be able to eliminate the unused branch. Plus we support older Julia versions as well, and the downsides to implementing `realdot` and `abs2` as we do are only needing to maintain 2 more LOCs, which is negligible.

hyrodium:
Thank you for the comments! This PR is ready for final review.

src/Quaternion.jl
```diff
@@ -386,3 +386,6 @@ LinearAlgebra.lyap(a::Quaternion, c::Quaternion) = lyap(promote(a, c)...)
 # if a commutes with c, use a simpler expression
 LinearAlgebra.lyap(a::Real, c::Quaternion) = c / -2a
 LinearAlgebra.lyap(a::Quaternion, c::Real) = c / -2real(a)
+
+## RealDot
+@inline RealDot.realdot(p::Quaternion, q::Quaternion) = (p.s * q.s + p.v2 * q.v2) + (p.v1 * q.v1 + p.v3 * q.v3)
```
sethaxen, line 391:
```suggestion
# ordering chosen so that real(p'q) == real(q'p) == realdot(p, q) == realdot(q, p), i.e. exact equality
@inline RealDot.realdot(p::Quaternion, q::Quaternion) = (p.s * q.s + p.v2 * q.v2) + (p.v1 * q.v1 + p.v3 * q.v3)
```

src/Quaternion.jl
```diff
@@ -321,7 +321,7 @@ true
     iszero(qb0) && throw(DomainError(qb0, "The input quaternion must be non-zero."))
     qa = qa0 / abs(qa0)
     qb = qb0 / abs(qb0)
-    coshalftheta = qa.s * qb.s + qa.v1 * qb.v1 + qa.v2 * qb.v2 + qa.v3 * qb.v3
+    coshalftheta = realdot(qa, qb)
```
sethaxen, line 324: I prefer we be explicit with functions not imported from the standard library.
```suggestion
    coshalftheta = RealDot.realdot(qa, qb)
```

test/Quaternion.jl
```diff
@@ -571,4 +571,13 @@ end
             @test_throws DivideError lyap(null, null)
         end
     end
+
+    @testset "RealDot with $T" for T in (Float32, Float64)
+        for _ in 1:10
+            q1 = randn(Quaternion{T})
+            q2 = randn(Quaternion{T})
+            @test real(dot(q1,q2)) == @inferred(realdot(q1,q2))
```
sethaxen, line 579: Can you also check the other 3 permutations? Something like this
```suggestion
            @test @inferred(realdot(q1,q2)) == realdot(q2, q1) == real(dot(q1, q2)) == real(dot(q2, q1))
```

sethaxen (changes requested):
I think we also said that we could use `realdot` for `abs2`?

hyrodium:
>I think we also said that we could use realdot for abs2?

Oh, sorry, I was forgetting that.

test/Quaternion.jl
hyrodium: Since checking `real(dot(q1, q2)) == real(dot(q2, q1))` does not seem to be our responsibility, I have added another commutative law test `@test realdot(q1,q2) == realdot(q2,q1)`. Is this okay?

sethaxen: Yes, that's fine, but I think we also want to check if one of those is equal to `real(dot(q2, q1))`.

hyrodium: I added the test in https://github.com/JuliaGeometry/Quaternions.jl/pull/119/commits/204e32079a8512db0d022a0f4b5706cc08df898e 👍

sethaxen (approved):
LGTM, thanks! Just please increment the patch version and make a patch release.

mikmoore (approved):

PR #120, opened 2022-12-10
fix typo in docstring

stevengj:
This seems to be a typo in the documentation (the code is correct).

From Sudbury:
image


sethaxen (approved):
LGTM, thanks for the fix!

PR #121, opened 2022-12-10
Implement more base methods

sethaxen:
I went through https://github.com/JuliaLang/julia/blob/v1.8.3/base/complex.jl and implemented here for `Quaternion` all overloads that are there implemented for `Complex`. All of these methods resulted in a `MethodError` before this PR, so this is a non-breaking change.

The one overload I did not implement is `hash`, which I would need to take some more time to understand.

### DIFF
--- a/test/Quaternion.jl
+++ b/test/Quaternion.jl
@@ -624,9 +624,8 @@ end
         @test round(q, RoundUp; digits=1) == q
         @test round(q, RoundUp, RoundToZero) == quat(2.0, 2.0, -3.0, 2.0)
         @test round(q, RoundUp, RoundToZero; digits=1) == q
-        @test round(q, RoundUp, RoundDown, RoundNearestTiesAway, RoundToZero) ==
-            quat(2.0, 2.0, -4.0, 2.0)
-        @test round(q, RoundUp, RoundDown, RoundNearestTiesAway, RoundToZero; digits=1) ==
-            q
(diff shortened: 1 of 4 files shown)


test/Quaternion.jl
```diff
@@ -571,4 +579,54 @@ end
...
+        @test round(q) == quat(1.0, 2.0, -4.0, 2.0)
+        @test round(q; digits=1) == q
+        @test round(q, RoundUp) == quat(2.0, 3.0, -3.0, 3.0)
+        @test round(q, RoundUp; digits=1) == q
+        @test round(q, RoundUp, RoundToZero) == quat(2.0, 2.0, -3.0, 2.0)
+        @test round(q, RoundUp, RoundToZero; digits=1) == q
+        @test round(q, RoundUp, RoundDown, RoundNearestTiesAway, RoundToZero) ==
+            quat(2.0, 2.0, -4.0, 2.0)
+        @test round(q, RoundUp, RoundDown, RoundNearestTiesAway, RoundToZero; digits=1) ==
+            q
```
hyrodium, line 630: I thought these line breaks can be dropped. If there is a limit to the number of characters per line, it is better to add `.JuliaFormatter.toml` file. (Adding JuliaFormatter can be separated into another PR.)

hyrodium (approved):
Added minor comments to the test code. The rest of the code seems fine:+1:

hyrodium (approved):
LGTM!

PR #124, opened 2023-01-09
Add CompatHelper

hyrodium:


sethaxen:
There are a few different CompatHelper workflows floating around, all of which seem to work but maybe have some failure modes. Where does this one come from? I noticed it doesn't match the ones in the CompatHelper repo https://github.com/JuliaRegistries/CompatHelper.jl/blob/master/.github/workflows/CompatHelper.yml or the PkgTemplates repo: https://github.com/JuliaCI/PkgTemplates.jl/blob/master/templates/github/workflows/CompatHelper.yml

hyrodium:
Oh sorry, my mistake. I copied this file from my other package. I thought the package has the latest `CompatHelper.yml` provided from CompatHelper.jl, but that was not correct.

sethaxen (approved):

PR #126, opened 2023-01-25
Update the version of Aqua

hyrodium:
CompatHelper does not check the `[compat]` table of versions of indirect dependencies, so I checked it:smile:

x-ref: https://github.com/JuliaRegistries/CompatHelper.jl/issues/247


sethaxen (approved):
Thanks!

PR #132, opened 2023-11-06
Add more methods to `quat`

hyrodium:
# Abstract
This PR adds more methods to `quat`.

**Before this PR**
```julia
julia> using Quaternions

julia> quat(missing)
ERROR: MethodError: no method matching Quaternion(::Missing)
```

**After this PR**
```julia
julia> using Quaternions

julia> quat(missing)
missing
```

This behavior is consistent with `Base.complex`.

```julia
julia> complex(missing)
missing
```

# How to find the methods to add
I have confirmed the methods with the `methods` function.
```julia
julia> using Quaternions

julia> methods(complex)
# 13 methods for generic function "complex" from Base:
  [1] complex(z::Complex)
     @ complex.jl:172
  [2] complex(x::Real)
     @ complex.jl:173
  [3] complex(x::Real, y::Real)
     @ complex.jl:174
  [4] complex(::Type{T}) where T<:Real
     @ complex.jl:191
  [5] complex(::Type{Complex{T}}) where T<:Real
     @ complex.jl:192
  [6] complex(S::SparseArrays.SparseMatrixCSC)
     @ SparseArrays ~/.julia/juliaup/julia-1.9.3+0.x64.linux.gnu/share/julia/stdlib/v1.9/SparseArrays/src/sparsematrix.jl:936
  [7] complex(x::SparseArrays.AbstractSparseVector{<:Complex})
     @ SparseArrays ~/.julia/juliaup/julia-1.9.3+0.x64.linux.gnu/share/julia/stdlib/v1.9/SparseArrays/src/sparsevector.jl:1083
  [8] complex(x::SparseArrays.AbstractSparseVector)
     @ SparseArrays ~/.julia/juliaup/julia-1.9.3+0.x64.linux.gnu/share/julia/stdlib/v1.9/SparseArrays/src/sparsevector.jl:1084
  [9] complex(A::AbstractArray{<:Complex})
     @ complex.jl:1117
 [10] complex(A::AbstractArray{T}) where T
     @ complex.jl:1119
 [11] complex(::Type{Missing})
     @ missing.jl:111
 [12] complex(::Type{Union{Missing, T}}) where T
     @ missing.jl:112
 [13] complex(::Missing)
     @ missing.jl:101

julia> methods(quat)
# 9 methods for generic function "quat" from Quaternions:
 [1] quat(q::Quaternion)
     @ ~/.julia/dev/Quaternions/src/Quaternion.jl:52
 [2] quat(s::Real)
     @ ~/.julia/dev/Quaternions/src/Quaternion.jl:53
 [3] quat(s::Real, v1::Real, v2::Real, v3::Real)
     @ ~/.julia/dev/Quaternions/src/Quaternion.jl:54
 [4] quat(A::AbstractArray{T}) where T
     @ ~/.julia/dev/Quaternions/src/Quaternion.jl:58
 [5] quat(::Type{T}) where T<:Real
     @ ~/.julia/dev/Quaternions/src/Quaternion.jl:80
 [6] quat(::Type{Quaternion{T}}) where T<:Real
     @ ~/.julia/dev/Quaternions/src/Quaternion.jl:81
 [7] quat(::Type{Missing})
     @ ~/.julia/dev/Quaternions/src/Quaternion.jl:85
 [8] quat(::Type{Union{Missing, T}}) where T
     @ ~/.julia/dev/Quaternions/src/Quaternion.jl:86
 [9] quat(::Missing)
     @ ~/.julia/dev/Quaternions/src/Quaternion.jl:83
```

Note that this PR does not contain methods with `SparseArrays` because there were some performance degressions.
```julia
julia> using BenchmarkTools, SparseArrays, Quaternions

julia> M = sprand(Float64, 16, 32, 0.5)
16×32 SparseMatrixCSC{Float64, Int64} with 261 stored entries:
⎡⣅⢶⡢⠫⠿⠢⢽⣃⠾⣠⣾⢕⠝⠅⠩⢗⎤
⎢⠃⢱⡜⣴⠓⣂⢼⡁⢴⣴⢘⡊⠯⡒⡚⣞⎥
⎢⣥⡕⠏⢯⠫⠲⠦⠓⢁⣂⢚⣹⣕⢺⢼⢎⎥
⎣⣜⡩⣛⢝⣄⢿⠈⠨⢴⢕⢒⡿⡡⣱⣑⡪⎦

julia> @benchmark quat(M)
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  1.300 μs … 208.680 μs  ┊ GC (min … max):  0.00% … 95.74%
 Time  (median):     1.831 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   2.394 μs ±   7.297 μs  ┊ GC (mean ± σ):  13.67% ±  4.51%

   ▂▄▄▅▆▇██▇▆▆▆▅▄▃▁▁                       ▁▂▂▂ ▁             ▂
  ▆███████████████████▇▇▇▇▆▆▆▄▅▅▅▂▄▅▅▂▅▇▇██████████████▇▇█▇▆▆ █
  1.3 μs       Histogram: log(frequency) by time      4.88 μs <

 Memory estimate: 10.88 KiB, allocs estimate: 4.

julia> Quaternions.quat(S::SparseMatrixCSC) = SparseMatrixCSC(size(S, 1), size(S, 2), copy(SparseArrays.getcolptr(S)), copy(rowvals(S)), quat(copy(nonzeros(S))))

julia> @benchmark quat(M)
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  1.631 μs … 180.029 μs  ┊ GC (min … max):  0.00% … 94.98%
 Time  (median):     2.170 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   2.681 μs ±   8.186 μs  ┊ GC (mean ± σ):  15.66% ±  5.05%

          ▃██▂                                                 
  ▁▂▂▂▂▃▄▇████▆▃▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  1.63 μs         Histogram: frequency by time        4.81 μs <

 Memory estimate: 13.06 KiB, allocs estimate: 5.
```

The above definition `Quaternions.quat(S::SparseMatrixCSC)` is from:

https://github.com/JuliaSparse/SparseArrays.jl/blob/bd2bda8f45ed71374223068b0a0daa2b169071c9/src/sparsematrix.jl#L984

### DIFF
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -65,7 +65,7 @@ end
 """
     quat(T::Type)
 
-Return an appropriate type which can represent a value of type `T` as a quaternion.
+Return an appropriate type that can represent a value of type `T` as a quaternion.
 Equivalent to `typeof(quat(zero(T)))`.
 
 # Examples
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -54,7 +54,7 @@ quat(s::Real) = Quaternion(s)
 quat(s::Real, v1::Real, v2::Real, v3::Real) = Quaternion(s, v1, v2, v3)
 
 ## Array operations on quaternions ##
-complex(A::AbstractArray{<:Complex}) = A
+quat(A::AbstractArray{<:Quaternion}) = A
 function quat(A::AbstractArray{T}) where T
     if !isconcretetype(T)
         error("`quat` not defined on abstractly-typed arrays; please convert to a more specific type")
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -28,9 +28,9 @@ Base.promote_rule(::Type{Quaternion{T}}, ::Type{S}) where {T <: Real, S <: Real}
 Base.promote_rule(::Type{Quaternion{T}}, ::Type{Quaternion{S}}) where {T <: Real, S <: Real} = Quaternion{promote_type(T, S)}
 
 """
-    quat(r, [i, j, k])
+    quat(w, [x, y, z])
 
-Convert real numbers or arrays to quaternion. `i, j, k` defaults to zero.
+Convert real numbers or arrays to quaternion. `x, y, z` defaults to zero.
 
 # Examples
 ```jldoctest
@@ -47,15 +47,45 @@ julia> quat([1, 2, 3])
  Quaternion{Int64}(3, 0, 0, 0)
 ```
 """
-quat
+function quat end
 
-quat(p, v1, v2, v3) = Quaternion(p, v1, v2, v3)
-quat(x) = Quaternion(x)
+quat(q::Quaternion) = q
+quat(s::Real) = Quaternion(s)
+quat(s::Real, v1::Real, v2::Real, v3::Real) = Quaternion(s, v1, v2, v3)
+
+## Array operations on quaternions ##
+complex(A::AbstractArray{<:Complex}) = A
 function quat(A::AbstractArray{T}) where T
     if !isconcretetype(T)
         error("`quat` not defined on abstractly-typed arrays; please convert to a more specific type")
     end
     convert(AbstractArray{typeof(quat(zero(T)))}, A)
+end
+
+"""
+    quat(T::Type)
+
+Return an appropriate type which can represent a value of type `T` as a quaternion.
+Equivalent to `typeof(quat(zero(T)))`.
+
+# Examples
+```jldoctest
+julia> quat(Quaternion{Int})
+Quaternion{Int64}
+
+julia> quat(Int)
+Quaternion{Int64}
+```
+"""
+quat(::Type{T}) where {T<:Real} = Quaternion{T}
+quat(::Type{Quaternion{T}}) where {T<:Real} = Quaternion{T}
+
+quat(::Missing) = missing
+# Same definitioin as in Base: https://github.com/JuliaLang/julia/blob/v1.9.3/base/missing.jl#L111-L115
+quat(::Type{Missing}) = Missing
+function quat(::Type{Union{T, Missing}}) where T
+    T === Any && throw(MethodError(quat, (Any,)))  # To prevent StackOverflowError
+    Union{quat(T), Missing}
 end
 
 """
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -47,7 +47,7 @@ julia> quat([1, 2, 3])
  Quaternion{Int64}(3, 0, 0, 0)
 ```
 """
-function quat end
+quat
 
 quat(q::Quaternion) = q
 quat(s::Real) = Quaternion(s)
--- a/test/Quaternion.jl
+++ b/test/Quaternion.jl
@@ -97,6 +97,8 @@ end
 
         @test quat(Int) === Quaternion{Int}
         @test quat(Float32) === Quaternion{Float32}
+        @test quat(Quaternion{Int}) === Quaternion{Int}
+        @test quat(Quaternion{Float32}) === Quaternion{Float32}
 
         # Note that `quat(1,missing,0,0)` throws an error.
         # This is the same behavior as `complex(1,missing)`.
--- a/test/Quaternion.jl
+++ b/test/Quaternion.jl
@@ -93,6 +93,16 @@ end
         @test quat(Quaternion(1, 2, 3, 4)) === Quaternion(1, 2, 3, 4)
         @test quat([2, 3, 4]) == Quaternion{Int}[2, 3, 4]
         @test_throws ErrorException quat(Real[1,2,3])
+        @test quat(Quaternion[1,2]) == Quaternion[1,2]
+
+        @test quat(Int) === Quaternion{Int}
+        @test quat(Float32) === Quaternion{Float32}
+
+        # Note that `quat(1,missing,0,0)` throws an error.
+        # This is the same behavior as `complex(1,missing)`.
+        @test quat(missing) === missing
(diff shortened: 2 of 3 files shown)


src/Quaternion.jl
```diff
@@ -47,17 +47,47 @@ julia> quat([1, 2, 3])
...
+quat(::Type{T}) where {T<:Real} = Quaternion{T}
+quat(::Type{Quaternion{T}}) where {T<:Real} = Quaternion{T}
+
+quat(::Missing) = missing
+# Same definitioin as in Base: https://github.com/JuliaLang/julia/blob/v1.9.3/base/missing.jl#L111-L115
+quat(::Type{Missing}) = Missing
+function quat(::Type{Union{T, Missing}}) where T
+    T === Any && throw(MethodError(quat, (Any,)))  # To prevent StackOverflowError
+    Union{quat(T), Missing}
+end
```
hyrodium, line 89: I don't know how to define this method without the unbounded type parameter.

src/Quaternion.jl
```diff
@@ -47,17 +47,47 @@ julia> quat([1, 2, 3])
  Quaternion{Int64}(3, 0, 0, 0)
 ```
 """
-quat
+function quat end
```
sethaxen, line 50: Why this change?

src/Quaternion.jl
```diff
@@ -47,17 +47,47 @@ julia> quat([1, 2, 3])
...
 
-quat(p, v1, v2, v3) = Quaternion(p, v1, v2, v3)
-quat(x) = Quaternion(x)
```
sethaxen, line 53: Technically this removal would be breaking. e.g. this would block users from constructing quaternions with `SymbolicUtils.Num` or `Unitful.Quantity`. Not that I'm opposed to it, but we need to be certain we're fine with that first.

src/Quaternion.jl
```diff
@@ -47,17 +47,47 @@ julia> quat([1, 2, 3])
...
     if !isconcretetype(T)
         error("`quat` not defined on abstractly-typed arrays; please convert to a more specific type")
     end
     convert(AbstractArray{typeof(quat(zero(T)))}, A)
 end
 
+"""
+    quat(T::Type)
+
+Return an appropriate type which can represent a value of type `T` as a quaternion.
```
sethaxen, line 68:
```suggestion
Return an appropriate type that can represent a value of type `T` as a quaternion.
```

src/Quaternion.jl
sethaxen: Yeah, it's a little ugly, but it's also what Base does, so for nor seems fine.

src/Quaternion.jl
hyrodium: I was thinking that explicit `funciton quat end` might be better for document generation, but it seems it does not make any difference. I will revert the change later.

src/Quaternion.jl
hyrodium: The docstring for `complex(::Type)` uses "which", but it seems "that" sounds more natural, so I will update as suggested.

https://github.com/JuliaLang/julia/blob/e7345b89fd4eb15e8f395395701e19be705d7b06/base/complex.jl#L180

src/Quaternion.jl
hyrodium: I think this is not a breaking change.

No problem with `Symbolics.Num` because it's a subtype of `Real`.

```julia
julia> using Symbolics, Quaternions

julia> a = Symbolics.Num(3)
3

julia> quat(a,a,a,a)
Quaternion{Num}(3, 3, 3, 3)

julia> quat(a)
Quaternion{Num}(3, 0, 0, 0)

julia> Symbolics.Num <: Real
true
```

`Unitful.Quantitiy` is already not compatible with Quaternions.jl.

**Before this PR**
```julia
julia> using Unitful, Quaternions

julia> d = 3u"°"
3°

julia> quat(d)
ERROR: MethodError: no method matching Quaternion(::Quantity{Int64, NoDims, Unitful.FreeUnits{(°,), NoDims, nothing}})

Closest candidates are:
  (::Type{T})(::T) where T<:Number
   @ Core boot.jl:792
  (::Type{T})(::AbstractChar) where T<:Union{AbstractChar, Number}
   @ Base char.jl:50
  (::Type{T})(::Base.TwicePrecision) where T<:Number
   @ Base twiceprecision.jl:266
  ...
```

**After this PR**
```julia
julia> using Unitful, Quaternions

julia> d = 3u"°"
3°

julia> quat(d)
ERROR: MethodError: no method matching quat(::Quantity{Int64, NoDims, Unitful.FreeUnits{(°,), NoDims, nothing}})

Closest candidates are:
  quat(::Quaternion)
   @ Quaternions ~/.julia/dev/Quaternions/src/Quaternion.jl:52
  quat(::Real)
   @ Quaternions ~/.julia/dev/Quaternions/src/Quaternion.jl:53
  quat(::Real, ::Real, ::Real, ::Real)
   @ Quaternions ~/.julia/dev/Quaternions/src/Quaternion.jl:54
  ...
```

src/Quaternion.jl
sethaxen: Yeah, "which" would convey a different in incorrect meaning here (also in Base).

sethaxen (approved):

PR #136, opened 2023-11-09
Add `Base.hash(::Quaternion, ::UInt)`

### LINKED ISSUE #135: `hash(::Quaternion, ::UInt)` should be defined
The following behavior is not consistent, and can be fixed by adding `hash(::Quaternion, ::UInt)` method.

```julia
julia> using Quaternions

julia> isequal(1, complex(1))
true

julia> 1 |> hash
0x5bca7c69b794f8ce

julia> complex(1) |> hash
0x5bca7c69b794f8ce

julia> isequal(1, quat(1))  # isequal(a,b) == true should imply `hash(a) == hash(b)`
true

julia> quat(1) |> hash
0x37cb992a8d339608

julia> unique(Complex[complex(2), complex(big"2")])
1-element Vector{Complex}:
 2 + 0im

julia> unique(Quaternion[quat(2), quat(big"2")])
2-element Vector{Quaternion}:
  Quaternion{Int64}(2, 0, 0, 0)
 Quaternion{BigInt}(2, 0, 0, 0)
```

x-ref: https://github.com/JuliaGeometry/Quaternions.jl/pull/121#issue-1488451695

hyrodium:
This PR fixes #135.

**Before this PR**
```julia
julia> using Quaternions

julia> isequal(1, complex(1))
true

julia> 1 |> hash
0x5bca7c69b794f8ce

julia> complex(1) |> hash
0x5bca7c69b794f8ce

julia> isequal(1, quat(1))  # isequal(a,b) == true should imply `hash(a) == hash(b)`
true

julia> quat(1) |> hash
0x37cb992a8d339608

julia> unique(Complex[complex(2), complex(big"2")])
1-element Vector{Complex}:
 2 + 0im

julia> unique(Quaternion[quat(2), quat(big"2")])
2-element Vector{Quaternion}:
  Quaternion{Int64}(2, 0, 0, 0)
 Quaternion{BigInt}(2, 0, 0, 0)
```

**After this PR**
```julia
julia> using Quaternions

julia> isequal(1, complex(1))
true

julia> 1 |> hash
0x5bca7c69b794f8ce

julia> complex(1) |> hash
0x5bca7c69b794f8ce

julia> isequal(1, quat(1))  # isequal(a,b) == true should imply `hash(a) == hash(b)`
true

julia> quat(1) |> hash
0x5bca7c69b794f8ce

julia> unique(Complex[complex(2), complex(big"2")])
1-element Vector{Complex}:
 2 + 0im

julia> unique(Quaternion[quat(2), quat(big"2")])
1-element Vector{Quaternion}:
 Quaternion{Int64}(2, 0, 0, 0)
```

### DIFF
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -244,7 +244,7 @@ end
 const hash_0_imags = hash.(0, h_imags)
 
 function Base.hash(z::Quaternion, h::UInt)
-    hash(real(z), h ⊻ ⊻(hash.(imag_part(z), h_imags)...) ⊻ ⊻(hash_0_imags...))
+    hash(real(z), h ⊻ xor(xor.(hash.(imag_part(z), h_imags), hash_0_imags)...))
 end
 
 """
--- a/src/Quaternion.jl
+++ b/src/Quaternion.jl
@@ -236,6 +236,17 @@ function Base.isequal(q::Quaternion, w::Quaternion)
     isequal(q.s, w.s) & isequal(q.v1, w.v1) & isequal(q.v2, w.v2) & isequal(q.v3, w.v3)
 end
 
+if UInt === UInt64
+    const h_imags = 0xfa29df508725c1bf, 0x5e6c4b26a400626d, 0x8dfd375bac9f9403
+else
+    const h_imags = 0x62802719, 0x5a9b072e, 0x209019b1
+end
+const hash_0_imags = hash.(0, h_imags)
+
+function Base.hash(z::Quaternion, h::UInt)
+    hash(real(z), h ⊻ ⊻(hash.(imag_part(z), h_imags)...) ⊻ ⊻(hash_0_imags...))
+end
+
 """
     extend_analytic(f, q::Quaternion)
 
--- a/test/Quaternion.jl
+++ b/test/Quaternion.jl
@@ -66,6 +66,16 @@ end
         @test !isequal(Quaternion(NaN, 0.0, Inf, -Inf), Quaternion(NaN, -0.0, Inf, -Inf))
     end
 
+    @testset "hash" begin
+        # https://github.com/JuliaGeometry/Quaternions.jl/issues/135
+        @test hash(0) == hash(quat(0)) == hash(0.0) == hash(quat(0.0))
+        @test hash(1) == hash(quat(1)) == hash(1.0) == hash(quat(1.0))
+        @test hash(quat(-0.0)) ≠ hash(quat(+0.0))
+        @test allunique([hash(quat(1,0,0,0)), hash(quat(0,1,0,0)), hash(quat(0,0,1,0)), hash(quat(0,0,0,1))])
+        @test hash(57, UInt(42)) == hash(quat(57), UInt(42))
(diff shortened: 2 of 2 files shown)


hyrodium:
Hmm, not sure why the Invalidation test failed..

image

https://github.com/JuliaGeometry/Quaternions.jl/actions/runs/6814371318/job/18531083622?pr=136

src/Quaternion.jl
```diff
@@ -236,6 +236,17 @@ function Base.isequal(q::Quaternion, w::Quaternion)
     isequal(q.s, w.s) & isequal(q.v1, w.v1) & isequal(q.v2, w.v2) & isequal(q.v3, w.v3)
 end
 
+if UInt === UInt64
```
sethaxen, line 239: Would this be better?
```suggestion
@static if UInt === UInt64
```

hyrodium: The `@static` macro brings better performance if it is used inside a function, but this case is not, so it seems there would not be better performance here.

```julia
julia> f() = @static Sys.iswindows() ? 1 : 2
f (generic function with 1 method)

julia> @code_lowered f()
CodeInfo(
1 ─     return 2
)

julia> g() = Sys.iswindows() ? 1 : 2
g (generic function with 1 method)

julia> @code_lowered g()
CodeInfo(
1 ─ %1 = Base.Sys.iswindows
│   %2 = (%1)()
└──      goto #3 if not %2
2 ─      return 1
3 ─      return 2
)
```

hyrodium:
@ranocha Do you know why the Invalidation check failed?

hyrodium:
The invalidation failure was because the number of `uinvalidated(invalidations)` has increased.

**Before this PR**
```julia
julia> using SnoopCompile, SnoopCompileCore

julia> invalidations = @snoopr begin using Quaternions end;

julia> uinvalidated(invalidations)
Set{Core.MethodInstance} with 1 element:
  MethodInstance for promote_rule(::Type, ::Type)

julia> inv_owned = length(filtermod(Quaternions, invalidation_trees(invalidations)))
0
```

**After this PR**
```julia
julia> using SnoopCompile, SnoopCompileCore

julia> invalidations = @snoopr begin using Quaternions end;

julia> uinvalidated(invalidations)
Set{Core.MethodInstance} with 2 elements:
  MethodInstance for hash(::Any, ::UInt64)
  MethodInstance for promote_rule(::Type, ::Type)

julia> inv_owned = length(filtermod(Quaternions, invalidation_trees(invalidations)))
0
```

But it seems this is not problematic because `inv_owned` is zero.

(The above code are from https://github.com/julia-actions/julia-invalidations/blob/main/action.yml.)

ranocha:
Based on your results above, it looks like there are some new invalidations caused by the new specialization. You don't invalidate your own functions but something that's using `hash` and is badly inferred. This may impact the latency when loading Quaternions.jl.
Of course, this doesn't mean that you cannot go ahead with this PR. It would be a nice contribution to the ecosystem to check where the invalidations are coming from and trying to fix them 🙂

hyrodium:
I see. Thank you for the detailed explanation!

hyrodium:
I'll merge this PR in a few days if there are no objections.

PR #138, opened 2023-11-25
Add CI test on v1.6

hyrodium:


sethaxen (approved):
This shouldn't be strictly necessary, since if the code works for 1.0 and the latest version it should also work for the LTS, but I don't think the extra safety is a bad thing.

